PR #365, opened 2019-01-16
Handle the case when an epigraph is inside a p

### LINKED ISSUE #362: Conversion of dtbook epigraph can cause a p inside a p
When converting dtbook to epub3 `epigraph`s are converted to `p`. This can cause an non-valid epub as you can end up with a `p` inside a `p`. You say that you do not use `dtb:epigraph` in the [Mapping between HTML and DTBook](http://nlbdev.github.io/nordic-epub3-dtbook-migrator/mapping.html), so this is probably not an issue for you.

Would you accept a PR that just emits an `aside` instead?

BTW, @josteinaj I'm not sure if filing issues is the right way to go about this. If you would like a different channel please tell me :-)

egli:
When an epigraph is used inside a p we should take care not to emit a
p inside a p as this is not permitted. Instead we enhance the
`has-block-elements` functionality of the `dtbook:p` matcher.

Fixes #362

### DIFF
--- a/src/main/resources/xml/xslt/dtbook-to-epub3.xsl
+++ b/src/main/resources/xml/xslt/dtbook-to-epub3.xsl
@@ -1119,7 +1119,7 @@
 
     <xsl:template match="dtbook:p">
         <xsl:variable name="element" select="."/>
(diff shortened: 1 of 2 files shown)


josteinaj:
Thanks. I'll merge along with #366 and #367 when I get some time.

egli:
This PR is now obsolete. It has been superseded by #371

PR #367, opened 2019-01-16
Render inline prodnotes as spans

### LINKED ISSUE #363: Conversion of dtbook prodnote can cause an aside inside a p
Converting an inline `prodnote` to EPUB3 can cause an `aside` inside a `p` which isn't valid according to epubcheck. 

Would you accept a PR that fixes this?

The `aside` would have to be moved before or after the current `p`. Do you have any idea how this could be achieved?

See also sbsdev/pipeline#7

egli:
Without this change inline prodnotes are rendered as asides. So you
might end up with non-valid EPUB as asides aren't allowed inside a
paragraph.

Fixes #363


josteinaj:
Merged in 1eea498db2ec71cfb1e386c7cc37d714519d5e43.

PR #371, opened 2019-01-22
Render dtbook:epigraph as html:blockquote

### LINKED ISSUE #368: Invalid HTML for dtbook:p inside dtbook:epigraph
The problem with epigraphs is not so much that we can have a `dtbook:epigraph` inside a `dtbook:p` (as described in #362 but that we can have `dtbook:p` inside `dtbook:epigraph`. This of course will also translate to `html:p` inside `html:p` which is not valid HTML5.

We propose to render `dtbook:epigraph` as `html:blockquote` as recommended by [WAI-ARIA](https://www.w3.org/TR/dpub-aria-1.0/#doc-epigraph). This will solve this issue and at the same time #362.

See also sbsdev/pipeline#11

egli:
This solves the issue that a p inside an epigraph can cause invalid
html. It also follows a recommendation by WAI-ARIA.

Fixes #368 

This PR replaces #365

### DIFF
--- a/src/test/xspec/dtbook-to-epub3.xspec
+++ b/src/test/xspec/dtbook-to-epub3.xspec
@@ -820,13 +820,34 @@
     <!-- implicit epub:type tested in element test, otherwise nothing interesting to test -->
 
     <!-- epigraph -->
+    <!-- The nordic guidelines do not allow for epigraph. So the
+         following is for completeness sake and takes some inspiration
+         from https://www.w3.org/TR/dpub-aria-1.0/#doc-epigraph -->
     <x:scenario label="when processing a epigraph element">
-        <x:context>
-            <dtbook:epigraph/>
-        </x:context>
-        <x:expect label="the resulting element should be a p with an added epub:type of epigraph">
-            <html:p id="..." epub:type="epigraph" class="epigraph"/>
-        </x:expect>
+	<x:scenario label="on its own">
+            <x:context>
+		<dtbook:epigraph/>
+            </x:context>
+            <x:expect label="the resulting element should be a blockquote with an added epub:type of epigraph">
+		<html:blockquote id="..." epub:type="epigraph" class="epigraph"/>
+            </x:expect>
+	</x:scenario>
+	<x:scenario label="which is inside a p">
+            <x:context>
+		<dtbook:p>foo<dtbook:epigraph/>bar</dtbook:p>
+            </x:context>
+            <x:expect label="the resulting element should be a p with a blockquote inside">
+	        <html:p>foo<html:blockquote id="..." epub:type="epigraph" class="epigraph"/>bar</html:p>
+            </x:expect>
+	</x:scenario>
+	<x:scenario label="which contains a p">
+            <x:context>
+		<dtbook:epigraph><dtbook:p>foo</dtbook:p></dtbook:epigraph>
+            </x:context>
+            <x:expect label="the resulting element should be a blockquote with a p inside">
+	        <html:blockquote id="..." epub:type="epigraph" class="epigraph"><html:p>foo</html:p></html:blockquote>
+            </x:expect>
+	</x:scenario>
     </x:scenario>
 
     <!-- attlist.epigraph -->
--- a/src/test/xspec/dtbook-to-epub3.xspec
+++ b/src/test/xspec/dtbook-to-epub3.xspec
@@ -832,14 +832,6 @@
 		<html:blockquote id="..." epub:type="epigraph" class="epigraph"/>
             </x:expect>
 	</x:scenario>
-	<x:scenario label="which is inside a p">
-            <x:context>
-		<dtbook:p>foo<dtbook:epigraph/>bar</dtbook:p>
-            </x:context>
-            <x:expect label="the resulting element should be a p with a blockquote inside">
-	        <html:p>foo<html:blockquote id="..." epub:type="epigraph" class="epigraph"/>bar</html:p>
-            </x:expect>
-	</x:scenario>
 	<x:scenario label="which contains a p">
             <x:context>
 		<dtbook:epigraph><dtbook:p>foo</dtbook:p></dtbook:epigraph>
--- a/src/main/resources/xml/schema/nordic-html5.rng
+++ b/src/main/resources/xml/schema/nordic-html5.rng
@@ -1787,7 +1787,7 @@
     or a division of a work.
   -->
     <define name="epigraph">
-        <element name="p">
+        <element name="blockquote">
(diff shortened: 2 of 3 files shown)


src/test/xspec/dtbook-to-epub3.xspec
```diff
@@ -820,13 +820,34 @@
...
+            <x:expect label="the resulting element should be a blockquote with an added epub:type of epigraph">
+		<html:blockquote id="..." epub:type="epigraph" class="epigraph"/>
+            </x:expect>
+	</x:scenario>
+	<x:scenario label="which is inside a p">
+            <x:context>
+		<dtbook:p>foo<dtbook:epigraph/>bar</dtbook:p>
+            </x:context>
+            <x:expect label="the resulting element should be a p with a blockquote inside">
+	        <html:p>foo<html:blockquote id="..." epub:type="epigraph" class="epigraph"/>bar</html:p>
```
josteinaj: According to `nordic-html5.rng`, `<p>` may only contain inline elements, and `<blockquote>` is a block element.

src/test/xspec/dtbook-to-epub3.xspec
```diff
@@ -820,13 +820,34 @@
...
-        </x:context>
-        <x:expect label="the resulting element should be a p with an added epub:type of epigraph">
-            <html:p id="..." epub:type="epigraph" class="epigraph"/>
-        </x:expect>
+	<x:scenario label="on its own">
+            <x:context>
+		<dtbook:epigraph/>
+            </x:context>
+            <x:expect label="the resulting element should be a blockquote with an added epub:type of epigraph">
+		<html:blockquote id="..." epub:type="epigraph" class="epigraph"/>
```
josteinaj: `nordic-html5.rng` defines epigraph as a `<p>`. The RelaxNG and Schematron should be updated where necessary.

src/test/xspec/dtbook-to-epub3.xspec
egli: Yes, agreed. Was a left over test from the previous PR

src/test/xspec/dtbook-to-epub3.xspec
josteinaj: ...also, in the RelaxNG/Schematron, maybe we should still allow `<p>` as epigraph in addition to `<blockquote>` so that existing books that are valid don't become invalid all of a sudden.

egli: So I changed the rng so that the epigraph should now use the element `blockquote`. I also changed the content model to flow as that seemed more appropriate.

I found no place in the schematron that I thought needed changing.

As far as backwards compatibility and also allowing `p`. I think this might be tricky especially since I'm also changing the content model. Since nobody was using `dtbook:epigraph` anyway I think we are on the safe side. And btw existing books were already invalid since they contain nested `html:p`.

egli: I'm not sure what to do about the comment about the HTML content model (see [nordic-html5.rng#L1791](https://github.com/nlbdev/nordic-epub3-dtbook-migrator/blob/master/src/main/resources/xml/schema/nordic-html5.rng#L1791)). What is that refering to? To `html:p`? So this would have to be changed to the content model of `html:blockquote`. Where did you get that information?

egli:
Jostein Austvik Jacobsen writes:

> Although, it's allowed to be used outside a wrapping `<p>` so it's
> *possible* to use it in a EPUB without validation errors.

I thought the problem was more that you can put a `dtbook:p` inside a
`dtbook:epigraph`. That will currently result in an `html:p` inside a
`html:p` and hence in invalid HTML.

> What do you think about this:
>
> ```xml
> <define name="epigraph">
>     <choice>
>         <element name="p"></element>
>         <element name="blockquote"></element>
>     </choice>
> </define>
> ```

That would work yes. But what about the content? ATM it says

```
<zeroOrMore>
    <ref name="flow"/> <!-- em, strong, dfn, code, samp, kbd, abbr, a, img, figure, br, q, sub, sup, span, bdo, p, ol, ul, dl, div, blockquote, table, address, section, (text) -->
</zeroOrMore>
```

Do you think it should instead be

```
<zeroOrMore>
    <choice>
      <ref name="inline"/> <!-- em, strong, dfn, code, samp, kbd, abbr, a, img, br, q, sub, sup, span, bdo, (text) -->
      <ref name="flow"/> <!-- em, strong, dfn, code, samp, kbd, abbr, a, img, figure, br, q, sub, sup, span, bdo, p, ol, ul, dl, div, blockquote, table, address, section, (text) -->
    </choice>
</zeroOrMore>
```

That seems too lax and a little bit of overkill if you ask me, but it's
your call :-)

> The comment about the HTML content model lists the relevant and
> allowed elements according to HTML5... You don't need to do anything
> about it.

OK, cool

-- 
Christian Egli
Swiss Library for the Blind, Visually Impaired and Print Disabled
Grubenstrasse 12, CH-8045 Zürich, Switzerland

egli:
Jostein Austvik Jacobsen writes:

> Although, it's allowed to be used outside a wrapping `<p>` so it's
> *possible* to use it in a EPUB without validation errors.

I thought the problem was more that you can put a `dtbook:p` inside a `dtbook:epigraph`. That will currently result in an `html:p` inside a `html:p` and hence in invalid HTML.

> What do you think about this:
>
> ```xml
> <define name="epigraph">
>     <choice>
>         <element name="p"></element>
>         <element name="blockquote"></element>
>     </choice>
> </define>
> ```

That would work yes. But what about the content? ATM it says

```xml
<zeroOrMore>
    <ref name="flow"/> <!-- em, strong, dfn, code, samp, kbd, abbr, a, img, figure, br, q, sub, sup, span, bdo, p, ol, ul, dl, div, blockquote, table, address, section, (text) -->
</zeroOrMore>
```

Do you think it should instead be

```xml
<zeroOrMore>
    <choice>
      <ref name="inline"/> <!-- em, strong, dfn, code, samp, kbd, abbr, a, img, br, q, sub, sup, span, bdo, (text) -->
      <ref name="flow"/> <!-- em, strong, dfn, code, samp, kbd, abbr, a, img, figure, br, q, sub, sup, span, bdo, p, ol, ul, dl, div, blockquote, table, address, section, (text) -->
    </choice>
</zeroOrMore>
```

That seems too lax and a little bit of overkill if you ask me, but it's your call :-)

> The comment about the HTML content model lists the relevant and allowed elements according to HTML5... You don't need to do anything about it.

OK, cool

josteinaj:
Thanks!

josteinaj:
Merged in b4c2ca253ff8a564d86ee864e199e38fbe0fe4bb...c7cc2edb1b38ed94ed4f583b0f391fcc9167fc5b.

PR #376, opened 2019-02-21
Map dtbook:poem to a html:div if it doesn't contain a headline

### LINKED ISSUE #374: Empty headers are added to poem without hd
When a book contains many poems we end up with many added headings, as the `dtbook:poem` are converted to `html:section` and [`generate-missing-headlines.xsl`](https://github.com/nlbdev/nordic-epub3-dtbook-migrator/blob/master/src/main/resources/xml/xslt/generate-missing-headlines.xsl) adds empty headlines to those. What aggravates the issue is that all of those end up in the navigation.

There seem to be two possible solutions to this:

### Convert poems without a heading to `html:div` instead of `html:section`
The [mapping](http://nlbdev.github.io/nordic-epub3-dtbook-migrator/mapping.html) says that a `dtbook:poem` without a header should be converted it to a `html:div`.

But in reality it converts a `dtbook:poem` without a header to a `html:section`. This is even confirmed by [a xspec test](https://github.com/nlbdev/nordic-epub3-dtbook-migrator/blob/master/src/test/xspec/dtbook-to-epub3.xspec#L930).

If poems were `html:div` then no empty headings would be added.

### Do not add headlines to poems
This is fairly easy to achieve, just exclude poems in [`generate-missing-headlines.xsl`](https://github.com/nlbdev/nordic-epub3-dtbook-migrator/blob/master/src/main/resources/xml/xslt/generate-missing-headlines.xsl)

Which way should a solution go?

egli:
According to the [mapping document](http://nlbdev.github.io/nordic-epub3-dtbook-migrator/mapping.html) a `dtbook:poem` should be converted to a `html:section` if it does contain a `dtbook:hd` and to a `html:div` otherwise.

Fixes #374

### DIFF
--- a/src/main/resources/xml/xslt/dtbook-to-epub3.xsl
+++ b/src/main/resources/xml/xslt/dtbook-to-epub3.xsl
@@ -700,10 +700,10 @@
     </xsl:template>
 
     <xsl:template match="dtbook:poem">
-        <section>
+        <xsl:element name="{if (dtbook:hd) then 'section' else 'div'}">
             <xsl:call-template name="f:attlist.poem"/>
             <xsl:apply-templates select="node()"/>
-        </section>
+	</xsl:element>
     </xsl:template>
 
     <xsl:template name="f:attlist.poem">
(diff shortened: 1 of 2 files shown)


josteinaj:
Merged in dd2c77690d11fa4ab391a845dd98cb62968eab85.

Issue #378, opened 2019-04-24
RuntimeException for long dtbook:li


egli:
When converting the attached DTBook file the converter crashes with the following exception:

```
INFO: 2019-04-23	14:39:04	172.17.0.1	-	172.17.0.2	8181	GET	/ws/jobs/36fd4926-b5b9-4f00-984e-949173085b1a	authid=clientid&time=2019-04-23T14%3A39%3A04Z&nonce=4sww%2BmGahDVA6oZmwEVpcba1gEBZlH8uXFgUvWyMwzs%3D&sign=cKRhyaeJgEqMDi7JhGIcTy%2BnNtk%3D	200	-	-	4	http://localhost:8181	Apache-HttpClient/4.5.5 (Java/1.8.0_191)	-
DEBUG  [     o.d.c.x.c.steps.Message$MessageStep] bundle://100.0:1/xml/xproc/message.xpl:116:29:Initialized in-memory fileset with xml:base="file:/opt/daisy-pipeline2/data/jobs/36fd4926-b5b9-4f00-984e-949173085b1a/context/"   @o.d.c.x.c.i.slf4jXProcMessageListener:91#fine
DEBUG  [     o.d.c.x.c.steps.Message$MessageStep] bundle://100.0:1/xml/xproc/message.xpl:116:29:processing file from memory: file:/opt/daisy-pipeline2/data/jobs/36fd4926-b5b9-4f00-984e-949173085b1a/context/EB10032.xml   @o.d.c.x.c.i.slf4jXProcMessageListener:91#fine
INFO   [     o.d.c.x.c.steps.Message$MessageStep] bundle://100.0:1/xml/xproc/message.xpl:116:29:DTBook preprocessing done   @o.d.c.x.c.i.slf4jXProcMessageListener:77#info
DEBUG  [     o.d.c.x.c.steps.Message$MessageStep] bundle://100.0:1/xml/xproc/message.xpl:116:29:Initialized in-memory fileset with xml:base="file:/opt/daisy-pipeline2/data/jobs/36fd4926-b5b9-4f00-984e-949173085b1a/context/"   @o.d.c.x.c.i.slf4jXProcMessageListener:91#fine
[...]

PR #379, opened 2019-04-24
Fix for "Cannot convert double nan to an integer"

egli:
When extracting the marker from a `dtbook:li` the regexp seems to run into some problems in some corner cases (probably new line combined with very long text inside the `dtbook:li`. This PR uses `substring-before` and should be more robust. All tests pass.

Fixes nlbdev/nordic-epub3-dtbook-migrator#378


josteinaj:
Merged in 2d8f81afb2b51edb8712262ecb519f662ad98924 and 73a692468ec435ec8c66c498e7c10402d79bd5e6.

PR #381, opened 2019-05-28
Add support for mailto and tel.

kalaspuffar:
I've only changed the epub_nordic_273b rule.

Sorry for the spacing changes. Hope these are ok.

### DIFF
--- a/src/main/resources/xml/schema/nordic2015-1.sch
+++ b/src/main/resources/xml/schema/nordic2015-1.sch
@@ -515,7 +515,7 @@
             <assert test="count(html:meta[@name='nordic:supplier'])=1">[nordic128d] nordic:supplier metadata must occur once.</assert>
         </rule>
     </pattern>
-    
+
     <!-- Rule 130 (44): dc:language must equal root element xml:lang -->
     <pattern id="epub_nordic_130">
         <rule context="html:meta[@name='dc:language']">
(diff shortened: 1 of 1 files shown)


josteinaj:
Looks good, thanks. I'll merge it in the next release.

josteinaj:
...although I think the regex should be `'^([a-z]+:/+|mailto:|tel:)'` and not `'^([a-z]+:/+)|mailto:|tel:'`? I'll have to check when I get back to a computer.

kalaspuffar:
Hi @josteinaj 

Well, it could be. It depends on the parser cohesion so we don't get a value that is or the character m and so on.
I can even see the possibility of using none capturing groups ```^(?:[a-z]+:/+)|(?:mailto:)|(?:tel:)``` in order to make it clear that we want these set of characters in a sequence and not a subset of them.

kalaspuffar:
Hi @josteinaj 

What is the status of this pull request? Still way from the computer?

Best regards
Daniel

josteinaj:
Hi @kalaspuffar.

I'm by the computer again :)

Making a complete new release of the migrator takes a little time. Not much, but some. And I'm swamped with other things right now. However, I merged it into the dev branch now, gave it a tag (in case there's more to come on the dev branch in the future), and set up Docker Hub to build such tags, so now it should very be available as the docker image `nlbdev/pipeline:2019-06-12`.

Are you using the docker image directly now?

PR #385, opened 2020-09-08
Make it work with Pipeline version 1.13.5 (modules-parent 1.13.6)

bertfrees:
MTM would like to use the migrator and prefers to not install a separate older version of Pipeline for it. @josteinaj would it be possible to make a new release with these changes?


bertfrees:
@josteinaj By the way, I've noticed that your installation instructions do not work anymore unless you use an old version of the updater. Some time ago I did this: https://github.com/daisy/pipeline-updater/commit/d19f3d94b57eb95c12fb9023cf79e8332c968bcc. This causes problems because the version defined in https://nlbdev.github.com/nordic-epub3-dtbook-migrator/releases/current is 1.4.2, which is not the version of the assembly, but the nordic-epub3-dtbook-migrator project.

josteinaj:
Hi Bert. Thanks for the PR!

I'm not able tor build the project though (see attached log): [modules-parent-1.13.6.log](https://github.com/nlbdev/nordic-epub3-dtbook-migrator/files/5200062/modules-parent-1.13.6.log)

There seems to be a problem with Maven Central now requiring HTTPS for everything. I'm guessing you've already got locally cached files so that it works for you.

When I get this to build, I can make a release of this very soon, but probably not this week. Hopefully next week.

The updater descriptor: when I've made a release, could you make a PR with the correct descriptor?

bertfrees:
> I'm not able tor build the project though

I don't think it's because I have it cached locally, but I have the correct repository set up in my local settings.xml file. You should do this too.

> The updater descriptor: when I've made a release, could you make a PR with the correct descriptor?

I'm not sure exactly how you create the descriptor. It should basically be the same as now, but I think the version should match the version of the main assembly. If different descriptors have no common versioning pattern, there is no way to verify whether one descriptor is compatible with another (i.e. whether an update is possible). What do you think?

josteinaj:
**build:**

I'm unable to build with Docker as well, which doesn't use my local settings.xml.

I believe this is the issue: https://issues.apache.org/jira/browse/MNG-5672

Is it required to have something special configured in settings.xml? I thought it was mainly if you want to push packages to sonatype.

**descriptor:**

It's been a while since I created it, I don't remember if it's generated automatically or manually, but for the last release I made, I just updated it manually.

If you use the updater to install the nordic migrator, then you essentially commit to using the nordic migrator version of Pipeline 2. There's no way to use multiple descriptors, or to have one descriptor extend another one, that I'm aware of. If there's a new version of the official Pipeline 2, then the nordic migrator descriptor will have to be updated as well.

The nordic migrator is not released along with the official version of Pipeline 2 so I'm not sure how a common versioning scheme would work. Would it be for instance `version="{official-version}-{nordic-version}"`, i.e. `version="1.13.5-1.4.2"`?

At NLB we just use the Docker image now, as it's much easier to set up. So I haven't tried using the updater in a while.

josteinaj:
By the way: I couldn't push to sbsdev, but I added a couple of commits to this branch that I'd like in a new release here:
https://github.com/nlbdev/nordic-epub3-dtbook-migrator/tree/modules-parent-1.13.6

- re-enabled tests
- updated healthcheck in Dockerfile

bertfrees:
> Is it required to have something special configured in settings.xml?

Yes. settings.xml can be used for various things, also setting up the repositories you want to fetch from.

If you don't want to touch the global settings file we could also add a settings file to the project. But for this case I think it is a good idea to use the global settings file because the HTTPS thing will affect pretty much every Maven project.

> If you use the updater to install the nordic migrator, then you essentially commit to using the nordic migrator version of Pipeline 2. There's no way to use multiple descriptors, or to have one descriptor extend another one, that I'm aware of. If there's a new version of the official Pipeline 2, then the nordic migrator descriptor will have to be updated as well.

What I mean is: there are at least two descriptors: the current one (official Pipeline version), and the nordic migrator one. The updater checks that the latter is compatible with the former, based on the two version numbers. At the moment it only checks that the new version is greater than the current version, but the check should be improved, e.g. to make sure the major and minor version numbers match.

The nordic migrator is not released along with the official version of Pipeline, but that doesn't mean that you should be able to update from any version of Pipeline. There needs to be a restriction. We decided that when there are changes that require a new download of the installer, we update the minor version number. So the updater should check this. This means that whenever we make a Pipeline release with a new minor version, a new release descriptor of the nordic migrator should be made.

`version="1.13.5-1.4.2"` could work. Maybe you should even add something to it that identifies the version as a "Nordic migrator" version. Although that might break the version compare code. But then again, that code needs some improvement anyway as I said before.

josteinaj:
Could you give an example of what is needed in the global settings.xml? We also need it within either the project POM or as a settings.xml in the project so that the project build in docker. It would also be good to have it in the project so that contributors can get up an running quickly.

bertfrees:
I have this is my global settings file:

```
    <profiles>
		<profile>
			<id>central-https</id>
			<repositories>
				<repository>
					<id>central-https</id>
					<url>https://repo.maven.apache.org/maven2</url>
					<releases>
						<enabled>true</enabled>
					</releases>
					<snapshots>
						<enabled>false</enabled>
					</snapshots>
				</repository>
			</repositories>
		</profile>
   	</profiles>
	<activeProfiles>
		<activeProfile>central-https</activeProfile>
	</activeProfiles>
```

We could also put this in a file named ".mvn/settings.xml" local to the project and add another file named ".mvn/maven.config" with the content

    --settings ./.mvn/settings.xml

This feature is described here: https://maven.apache.org/configure.html#mvn-maven-config-file.

src/test/xprocspec/dtbook-to-html.step.xprocspec
```diff
@@ -28,7 +28,7 @@
         <x:expect type="xpath" label="The output document (the intermediary HTML) should have an xml:base attribute with filename C00000.xhtml" test="tokenize(/*/@xml:base,'/')[last()]"
             equals="'C00000.xhtml'"/>
         <x:expect type="xpath" label="The output document (the intermediary HTML) should have an xml:base attribute pointing to the temp-dir directory" test="/*/@xml:base"
-            equals="concat($temp-dir,'temp-dir/C00000.xhtml')"/>
+            equals="concat($temp-dir,'temp-dir/C00000.xhtml')" pending="the base URI is normalized"/>
```
josteinaj, line 31: Ah, I see now that this is also in this PR, not only in #387. I have the same question here though: why is it pending? Does the xml:base not contain the full path anymore?

josteinaj:
Thanks, I have it working now. `.mvn/settings.xml` didn't have any effect so I used the Dockerfile to move it to `~/.m2/settings.xml` inside the docker image, which worked.

So I suppose I can make a release of this now, and make a new updater descriptor based on 1.13.5?
https://github.com/daisy/pipeline-assembly/blob/gh-pages/releases/1.13.5

bertfrees:
That would be great!

Still not sure what the version attribute should be. Basically anything that is higher than 1.13.5 will do. "1.13.5-p1". Or "1.13.5-nordic-epub3-dtbook-migrator-1.4.4".

> .mvn/settings.xml didn't have any effect

Hmm, too bad. Are you sure you also created the .mvn/maven.config file? Are you using a recent version of Maven? I think this is a feature that doesn't exist that long yet.

josteinaj:
Yes, I tried with `maven.config`.

My laptop:
```zsh
➜  ~ mvn --version
Apache Maven 3.6.3
```

Dockerfile:
```Dockerfile
FROM maven:3.6-jdk-11 as builder
```

bertfrees:
Same version as I have, and with that version it is supposed to work. I didn't try it myself though. Oh well, at least you got it working with Docker.

src/test/xprocspec/dtbook-to-html.step.xprocspec
bertfrees: I don't remember exactly what the URI was that I was getting. It was equivalent (normalized) but could not be compared the way it was done. However I tested it again now and it seems the pending is not needed. Not sure what happened. Can you try to remove it and run the test again?

josteinaj: It works for me as well. :+1:

josteinaj:
Replaced by: #388.

josteinaj:
I've tested that it works with the Linux ZIP of Pipeline 2 v1.13.0 and then using the updater. The docker image is still building, but it built fine locally so I excect that will be done any minute now as well.

bertfrees:
Brilliant, thanks!

PR #411, opened 2021-02-08
Removing rules mentioned in the review as removable.

kalaspuffar:
Hi @josteinaj 

This is my first PR with the removal of some of the rules mentioned as removable in the conversation. I hope this is a good workflow going forward.

Best regards
Daniel


josteinaj:
Thanks. Looks good. Yes, this is a good workflow.

PR #412, opened 2021-02-10
Cleanup

AlexanderHugestrand:
Hi @josteinaj, 

I went through remaining issues regarding rules 51-100, but most seemed to be completed already. So I have only removed trailing whitespaces, a few TODOs, and adjusted one condition. I don't feel confident in any of those changes, so please update me with your thoughts and philosophy regarding this.

Best regards,

Alexander

### DIFF
--- a/src/main/resources/xml/schema/2020-1/nordic2020-1.sch
+++ b/src/main/resources/xml/schema/2020-1/nordic2020-1.sch
@@ -319,7 +319,7 @@
         <rule context="html:figure[tokenize(@class, '\s+') = 'image']">
             <let name="context" value="concat('(&lt;', name(), string-join(for $a in (@*) return concat(' ', $a/name(), '=&quot;', $a, '&quot;'), ''), '&gt;)')"/>
             <assert test="html:img">[nordic102] There must be an img element in every figure with class="image". <value-of select="$context"/></assert>
-            <report test="parent::html:figure">[nordic102] Nested image figures are not allowed. Image figures must use the class "image", while image series figures must use the class "image-series". <value-of select="$context"/></report>
+            <report test="parent::html:figure">[nordic102] Figures within image figures not allowed. Image figures must use the class "image", while image series figures must use the class "image-series". <value-of select="$context"/></report>
         </rule>
     </pattern>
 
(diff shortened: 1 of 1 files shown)


src/main/resources/xml/schema/2020-1/nordic2020-1.sch
```diff
@@ -314,18 +312,17 @@
...
-    <!-- TODO: Depending on https://github.com/nlbdev/epub3-guidelines-update/issues/98 -->
+
     <pattern id="epub_nordic_102">
         <title>Rule 102</title>
         <p>All image figures must have a image</p>
         <rule context="html:figure[tokenize(@class, '\s+') = 'image']">
             <let name="context" value="concat('(&lt;', name(), string-join(for $a in (@*) return concat(' ', $a/name(), '=&quot;', $a, '&quot;'), ''), '&gt;)')"/>
             <assert test="html:img">[nordic102] There must be an img element in every figure with class="image". <value-of select="$context"/></assert>
-            <report test="parent::html:figure[tokenize(@class, '\s+') = 'image']">[nordic102] Nested images figures are not allowed. Image figures must use the class "image", while image series figures must use the class "image-series". <value-of select="$context"/></report>
+            <report test="parent::html:figure">[nordic102] Figures within image figures not allowed. Image figures must use the class "image", while image series figures must use the class "image-series". <value-of select="$context"/></report>
```
josteinaj, line 322: @AlexanderHugestrand maybe this should be checked with @martinpub and @AndersEklund. It is probably correct but it makes the rule more generic as well. Could there in theory be cases where figures - except image figures - should be allowed?

josteinaj (commented):
Looks good. Should maybe get some feedback from @martinpub or @AndersEklund about the figures-rule as well.

src/main/resources/xml/schema/2020-1/nordic2020-1.sch
AlexanderHugestrand: @martinpub and @Anderseklund, can I please have your input on this?

martinpub: Sorry for the delayed response here. I would say: Nested figures with class image should probably not be allowed, it doesn't make sense given the image-series vs. image class attribute contsruction. We do want to allow structures like `<figure class ="table"><table> [...]`, however (and this is what MTM will use for grouping tables and table descriptions together), but I don't think this rule blocks that, right?

PR #413, opened 2021-02-10
Fixing rule with epub:type not allowed for body element

kalaspuffar:
Hi @josteinaj 

I've now changed this PR only to have the epub:type fix.

The aria rules will either be checked by ACE or be added in a later revision per our meeting.

Best regards
Daniel


josteinaj:
Looks good.

PR #416, opened 2021-02-22
Allow headers in nav-ncx

kalaspuffar:
Hi @josteinaj 

One rule that needed changing for the NAV-NCX document was to allow h[x] elements in the nav document. I've looked into the EPUB standard, and it says that only one header element is allowed in a nav element, so it should be fine to relax this rule a bit and allow the other header elements as well.

https://www.w3.org/publishing/epub3/epub-packages.html#sec-package-nav-def

Then again, I don't see a reason to use another header element for a TOC or page list header.

Best regards
Daniel


josteinaj:
This seems fine (unless @martinpub / @AndersEkl has any objections).

I think you forgot to select `guidelines-revision` as the target branch.

kalaspuffar:
Hi @josteinaj 

This change comes from the comment of @martinpub 

https://github.com/nlbdev/epub3-guidelines-update/issues/83#issuecomment-756086967

Best regards
Daniel

martinpub:
There could be cases where the nav document is included in the spine, and that might call for another heading structure to be used in the nav elements. However, the guidelines currently propose <h1> for all three nav elements (toc, page list, landmarks), so it's not an issue. Having the rule slightly relaxed is still a good idea I think.

PR #418, opened 2021-02-22
Nav references update

kalaspuffar:
Hi @josteinaj 

I've now done the requested changes for the nav reference document.

The changes identified:
* [nordic_nav_references_2a]: Change to require that references are made to sectioning elements only (not headings)
* [nordic_nav_references_3]:3: Remove, no pagebreaks will be without a page value.

So we don't check references to headings only the sectioning items and page number without a value is not allowed and will be handled by the agencies producing the material.

Best regards
Daniel


josteinaj:
Looks good.

PR #421, opened 2021-02-23
Revert "Nav references update"

kalaspuffar:
Reverts nlbdev/nordic-epub3-dtbook-migrator#418

Hi @josteinaj 

I was changing another rule and was wondering why this change had not been applied yet. And realized that I'd updated the old rules. This needs to be reverted. Will create a new PR for the new ruleset.

Best regards
Daniel


josteinaj:
Aha, right, yeah I should remember to check the file path before accepting PRs as well :). Good thing you spotted it.

PR #424, opened 2021-02-23
Making changes to handle the first 6 comments on the content structure.

kalaspuffar:
Hi @josteinaj 

I've started to look into the RelaxNG files today and update the rules after the comments made on GitHub. Some of them are quite broad so I might have misunderstood the range of the change.

1. define name="book" is not descriptive in a multipage html validation setting
2. top-level section is mandatory and should have the semantic attributes currently placed on body, body should not have any semantic attributes
3. attribute role should be allowed in every location where epub:type is currently allowed (however, there is not a 1-1 relationship between epub:type and role, in some places a matching role is not available, so it shouldn't be always required)
4. types: remove all z3998 values? No, rather make them optional!
5. types: update epub:type list based on https://idpf.github.io/epub-vocabs/structure/ (e.g. rearnotes are deprecated in favour of endnotes)
6. level1.content: currently h1 is the only allowed top-level heading, at least h2 should be allowed too, given the fact that part sectioning will have chapters starting on h2

Best regards
Daniel

### DIFF
--- a/src/main/resources/xml/schema/2020-1/nordic-html5.rng
+++ b/src/main/resources/xml/schema/2020-1/nordic-html5.rng
@@ -909,8 +909,14 @@
         <ref name="attlist.level1"/> <!-- @id, @title, @xml:space, @xml:lang, @lang, @dir, @class -->
         <choice>
             <group>
-                <ref name="h1"/> <!-- h1 -->
-                <ref name="h2"/> <!-- h2 -->
+                <choice>
+                    <ref name="h1"/> <!-- h1 -->
+                    <ref name="h2"/> <!-- h2 -->
+                    <ref name="h3"/> <!-- h3 -->
+                    <ref name="h4"/> <!-- h4 -->
+                    <ref name="h5"/> <!-- h5 -->
+                    <ref name="h6"/> <!-- h6 -->
+                </choice>
                 <oneOrMore>
                     <choice>
                         <ref name="docblockorinline"/> <!-- p, hr, ol, ul, dl, pre, div, blockquote, img, figure, section, span, table, address, aside, math -->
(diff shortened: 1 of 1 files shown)


src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -927,6 +910,7 @@
         <choice>
             <group>
                 <ref name="h1"/> <!-- h1 -->
+                <ref name="h2"/> <!-- h2 -->
```
josteinaj, line 913: This will require both a h1 and a h2. Use a `choice` to require either or:

```xml
<choice>
    <group>
        <choice>
            <ref name="h1"/> <!-- h1 -->
            <ref name="h2"/> <!-- h1 -->
        </choice>
        <oneOrMore>
```
```

**Review by josteinaj — CHANGES_REQUESTED** [review]
Yes that's correct.

For allowing h2 as well as h1 you need a `<choice>`.

Also, if using a h2, then the next level must use a h3 instead of h2, and the next level after that must use h4 instead of h3, etcetera. Maybe we should just allow all heading elements, and instead validate the level numbers in schematron?

File: src/main/resources/xml/schema/2020-1/nordic-html5.rng [hunk 2]
```diff
@@ -943,6 +927,7 @@
                 </oneOrMore>
                 <optional>
                     <ref name="h1"/> <!-- h1 -->
+                    <ref name="h2"/> <!-- h2 -->
```
josteinaj, line 930: (same here)

kalaspuffar:
Hi @josteinaj 

Added a choice and changed to all available heading levels.

Best regards
Daniel

josteinaj:
Looks good. Thanks!

PR #425, opened 2021-02-23
Batch 250 plus

AlexanderHugestrand:
@josteinaj

### DIFF
--- a/src/main/resources/xml/schema/2020-1/nordic2020-1.sch
+++ b/src/main/resources/xml/schema/2020-1/nordic2020-1.sch
@@ -764,7 +764,7 @@
                     else
                         ()"/>
             <assert test="
-                    not(matches(base-uri(.), $filename-regex)) or (for $t in tokenize(@epub:type, '\s+')
+                    not(matches(base-uri(.), $filename-regex)) or (for $t in tokenize(@role, '\s+')
                     return tokenize($t, ':')[last()]) = $base-uri-type">[nordic269] The type used in the filename (<value-of select="$base-uri-type"/>) must be present on the section element. <value-of select="$context"/></assert>
         </rule>
     </pattern>
--- a/src/main/resources/xml/schema/2020-1/nordic2020-1.sch
+++ b/src/main/resources/xml/schema/2020-1/nordic2020-1.sch
@@ -595,11 +595,15 @@
         <title>Rule 253a</title>
         <p>figures and captions</p>
         <rule context="html:figure">
+            <!--
+                The class "table" is undocumented to avoid unwanted use of that markup.
+                It should only be used by suppliers explicitly instructed to do so.
+            -->
             <let name="context" value="concat('(&lt;', name(), string-join(for $a in (@*) return concat(' ', $a/name(), '=&quot;', $a, '&quot;'), ''), '&gt;)')"/>
-            <assert test="tokenize(@epub:type, '\s+') = 'sidebar' or tokenize(@class, '\s+') = ('image', 'image-series')"
-                >[nordic253a] &lt;figure&gt; elements must either have an epub:type of "sidebar" or a class of "image" or "image-series". <value-of select="$context"/></assert>
-            <report test="count((.[tokenize(@epub:type, '\s+') = 'sidebar'], .[tokenize(@class, '\s+') = 'image'], .[tokenize(@class, '\s+') = 'image-series'])) &gt; 1"
-                >[nordic253a] &lt;figure&gt; elements must either have an epub:type of "sidebar" or a class of "image" or "image-series". <value-of select="$context"/></report>
+            <assert test="tokenize(@class, '\s+') = ('image', 'image-series', 'table')"
+                >[nordic253a] &lt;figure&gt; elements must either have a class of "image", "image-series" or "table". <value-of select="$context"/></assert>
+            <report test="count((.[tokenize(@class, '\s+') = 'image'], .[tokenize(@class, '\s+') = 'image-series'], .[tokenize(@class, '\s+') = 'table'])) &gt; 1"
+                >[nordic253a] &lt;figure&gt; elements must either have a class of "image", "image-series" or "table". <value-of select="$context"/></report>
             <assert test="count(html:figcaption) &lt;= 1">[nordic253a] There cannot be more than one &lt;figcaption&gt; in a single figure element. <value-of select="$context"/></assert>
         </rule>
     </pattern>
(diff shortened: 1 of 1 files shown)


src/main/resources/xml/schema/2020-1/nordic2020-1.sch
```diff
@@ -784,7 +779,7 @@
     <pattern id="epub_nordic_269">
         <title>Rule 269</title>
         <p></p>
-        <rule context="html:body[not(html:header)]">
+        <rule context="html:body/html:section[not(html:header)]">
```
josteinaj, line 782: Note that checking for the header elment is not necessary any more. The `<header>` was used in the 2015-1 guidelines as a wrapper around the HTML elements corresponding to [the DTBook `<doctitle>`, `<docauthor>` and `covertitle` elements](https://github.com/daisy/pipeline/blob/f65e446baa8764d85b5a4f26a3e88dff5e786191/modules/scripts-utils/dtbook-utils/src/main/resources/xml/dtd/dtbook-2005-3.dtd#L2209-L2241). The `<header>` element was the only allowed element other than `<section>` (and the never-used `<article>`), and was always always present in the HTML version (single HTML file), but never present in the EPUB version (multiple HTML files). There are several places in the schematrons where we check for the header element, and it is almost always used as a test for whether we're validating a single HTML file, or a HTML file which is part of an EPUB.

In this case, the context `html:body[not(html:header)]` matches `<body>` elements for when the book is split into multiple HTML files (EPUB) but not when the book is in a single HTML file (HTML version).

src/main/resources/xml/schema/2020-1/nordic2020-1.sch
```diff
@@ -794,7 +789,7 @@
                         ()"/>
             <assert test="
                     not(matches(base-uri(.), $filename-regex)) or (for $t in tokenize(@epub:type, '\s+')
```
josteinaj, line 791: This rule has changed in the new guidelines. It is now based on `role` instead of `epub:type`. See: https://format.mtm.se/nordic_epub/2020-1/#file-naming-convention

josteinaj (changes requested):

src/main/resources/xml/schema/2020-1/nordic2020-1.sch
AlexanderHugestrand: I have removed it. Please check the update.

src/main/resources/xml/schema/2020-1/nordic2020-1.sch
AlexanderHugestrand: I have replaced epub:type with role now.

src/main/resources/xml/schema/2020-1/nordic2020-1.sch
```diff
@@ -764,7 +764,7 @@
                     else
                         ()"/>
             <assert test="
-                    not(matches(base-uri(.), $filename-regex)) or (for $t in tokenize(@epub:type, '\s+')
+                    not(matches(base-uri(.), $filename-regex)) or (for $t in tokenize(@role, '\s+')
                     return tokenize($t, ':')[last()]) = $base-uri-type">[nordic269] The type used in the filename (<value-of select="$base-uri-type"/>) must be present on the section element. <value-of select="$context"/></assert>
```
josteinaj, line 768: After checking the guidelines, I see that some improvements are still needed here. Full description of file naming: https://github.com/nlbdev/epub3-guidelines-update/blob/master/guidelines/guidelines.md#file-naming-convention

1. _role_ corresponds to the ARIA `role` of the main section element in the content file, **minus the "`doc-`" part.**
2. In the absence of an ARIA role for the top-level section of the content document, the value from the `epub:type` attribute should be used instead.
3. In the case of multiple `epub:type` (e.g. `frontmatter titlepage`) values, the most specific (e.g. `titlepage`) should be used.

### Suggestion (untested):

First define a couple of useful variables:

```
<!-- from https://idpf.github.io/epub-vocabs/structure/ -->
<let name="document-partitions" select="('cover', 'frontmatter', 'bodymatter', 'backmatter')"/>
<let name="document-divisions" select="('volume', 'part', 'chapter', 'division')"/>
```

Then update the XPath in `test`:
```
not(matches(base-uri(.), $filename-regex))
not(matches(base-uri(.), $filename-regex))
or (
    (for $t in tokenize(@role, '\s+') return tokenize(replace($t, '^doc-', ''), ':')),
    (for $t in tokenize(@epub:type, '\s+') return if $t = ($document-partitions, $document-divisions) then () else tokenize($t, ':')),
    (for $t in tokenize(@epub:type, '\s+') return if $t = $document-divisions then $t else ()),
    (for $t in tokenize(@epub:type, '\s+') return if $t = $document-partitions then $t else ()),
)[1] = $base-uri-type
```

The first for loop creates a sequence of all the roles, and stripping any `doc-` prefixes. The second for loop creates a list of all the epub:types except for the types used as document partitions or divisions (according to the [EPUB 3 Structural Semantics Vocabulary](https://idpf.github.io/epub-vocabs/structure/)). The third loop creates a list of all the document division types, and the last for loop creates a list of all the document partition types. The surrounding parenthesis creates a list of all these in the preferred order (prefer role, fall back to most specific epub:type). The `[1]` after the parenthesis selects the preferred type (previously it was `[last()]` which assumed that the most specific type was declared last, such as `bodymatter chapter`, but it might as well be declared the other way around: `chapter bodymatter`).

josteinaj (changes requested):

src/main/resources/xml/schema/2020-1/nordic2020-1.sch
AlexanderHugestrand: @josteinaj What's the idea of calling tokenize() with ':' in the first for-loop?

josteinaj: It's mainly for the z3998 vocabulary: https://www.daisy.org/z3998/2012/vocab/structure/

Types that are not defined in the official EPUB vocabulary can be included if their prefix is declared using the [`epub:prefix`](https://www.w3.org/publishing/epub/epub-packages.html#sec-prefix-attr) attribute on the `<html>` element. We've been using the z3998 vocabulary for the guidelines since it is more rich than the official vocabulary. We didn't require the `epub:prefix` attribute to be used in the previous guidelines, although we should have. You don't need to check it for this validation rule, but for reference, this is how the `z3998` prefix is declared:

```html
<html epub:prefix="z3998: http://www.daisy.org/z3998/2012/vocab/structure/#">
```

When using types from that vocabulary, we prefix them with "z3998". For instance:

```html
<section epub:type="frontmatter z3998:grant-acknowledgment">
```

As we don't want to include the `z3998`-prefix in the file names, we have `tokenize($t, ':')[last()]` to get everything after the colon, if there's a colon. We could also use `replace($t, '.*:', '')` if that's more readable. Or more explicitly: `if (contains($t, ':')) then substring-after($t, ':') else $t`.

AlexanderHugestrand: @josteinaj Thanks for the explanation! I have tested your solution and it seems to work. Please take a look at the latest changes. I assigned the result of those for-loops to a variable to be able to reuse them in the error message.

josteinaj (approved):

josteinaj:
Looks great, thanks!

PR #426, opened 2021-02-23
Fixes for batch 2.

kalaspuffar:
Hi @josteinaj 

Another batch of relaxng rules.

* epub:type=sidebar is deprecated. Figure should allow other content than images, even when not used as sidebar.
* figcaption should now precede the img element in figure
* longdesc is deprecated, remove
* Preformatted lists now have the class attribute "plain", and should be allowed also for `<ul>`s

Best regards
Daniel


josteinaj:
Looks good.

PR #427, opened 2021-02-23
Batch 3 fixes.

kalaspuffar:
Hi @josteinaj 

* Remove class attribute notebody
* aside should be used for body footnotes (one aside element per footnote) only, endnotes should be in a list along the lines set out by http://kb.daisy.org/publishing/docs/html/notes.html
* Update poem section to reflect new guidelines

New guidelines for poems

* Poetry, song lyrics or any content written in verse, where lines of text must be preserved just as they are in the source material, is required to be marked up with ```<div class="verse">```.
* Each group of lines of text must be contained in a separate ```<p class="linegroup">``` element and each line of text must be marked up with ```<span class="line">```. HTML line breaks, ```<br/>```, must be added between consecutive lines within a line group. Indented lines may be marked up using ```<span class="line_indent">``` or ```<span class="line_longindent">.```
*Line numbers must only be included if specific instructions are given about it, even if they are present in the source material. If line numbers are to be included, they must be marked up with ```<span class="linenum">```.
* If the content written in verse has a title it may be handled as a normal heading, with ```<section>``` elements wrapping the content. However, if it does not make sense to use a heading, it may be marked up with ```<p class="verse-title">``` and placed within the ```<div class="verse">``` container. This option must not be used unless specific instructions are given by the Ordering Agency.
* If there is an author name placed under the verse it may be marked up with ```<p class="verse-author">``` and placed at the end of the ```<div class="verse">``` container.

Best regards
Daniel

### DIFF
--- a/src/main/resources/xml/schema/2020-1/nordic-html5.rng
+++ b/src/main/resources/xml/schema/2020-1/nordic-html5.rng
@@ -2095,21 +2095,28 @@
             <attribute name="class">
                 <value>verse</value>
             </attribute>
-            <optional>
+            <zeroOrMore>
                 <ref name="bridgehead"/>
-            </optional>
+                <ref name="poem.author"/>
+            </zeroOrMore>
             <oneOrMore>
                 <ref name="poem.linegroup"/>
             </oneOrMore>
-            <optional>
-                <element name="p">
-                    <attribute name="class">
-                        <value>verse-author</value>
-                    </attribute>
-                </element>
-            </optional>
+            <zeroOrMore>
+                <ref name="bridgehead"/>
+                <ref name="poem.author"/>
+            </zeroOrMore>
         </element>
     </define>
+
+    <define name="poem.author">
+        <element name="p">
+            <attribute name="class">
+                <value>verse-author</value>
+            </attribute>
+        </element>
+    </define>
+
 
     <define name="poem.linegroup">
         <element name="p">
@@ -2126,11 +2133,11 @@
     <define name="poem.line">
         <element name="span">
             <attribute name="class">
-                <choice>
+                <oneOrMore>
                     <value>line</value>
                     <value>line_indent</value>
                     <value>line_longindent</value>
-                </choice>
+                </oneOrMore>
             </attribute>
             <zeroOrMore>
                 <ref name="inline"/> <!-- em, strong, dfn, code, samp, kbd, span, abbr, math, a, img, br, q, sub, sup, bdo -->
--- a/src/main/resources/xml/schema/2020-1/nordic-html5.rng
+++ b/src/main/resources/xml/schema/2020-1/nordic-html5.rng
@@ -1866,16 +1866,17 @@
         <attribute name="epub:type">
             <list>
                 <ref name="types"/>
-                <value>note</value>
                 <value>footnote</value>
                 <ref name="types"/>
             </list>
         </attribute>
-        <attribute name="class">
-            <list>
-                <ref name="classes"/>
-            </list>
-        </attribute>
+        <optional>
+            <attribute name="class">
+                <list>
+                    <ref name="classes"/>
+                </list>
+            </attribute>
+        </optional>
         <ref name="attrsrqd.base"/> <!-- @id, @title, @xml:space, @xml:lang, @lang, @dir -->
     </define>
     
@@ -2095,12 +2096,11 @@
                 <value>verse</value>
             </attribute>
             <optional>
-                <choice>
-                    <ref name="title"/> <!-- h1, h2, h3, h4, h5, h6 -->
-                    <ref name="bridgehead"/>
-                </choice>
+                <ref name="bridgehead"/>
             </optional>
-            <ref name="poem.linegroup"/>
+            <oneOrMore>
+                <ref name="poem.linegroup"/>
+            </oneOrMore>
             <optional>
                 <element name="p">
                     <attribute name="class">
--- a/src/main/resources/xml/schema/2020-1/nordic-html5.rng
+++ b/src/main/resources/xml/schema/2020-1/nordic-html5.rng
@@ -1867,13 +1867,12 @@
             <list>
                 <ref name="types"/>
                 <value>note</value>
+                <value>footnote</value>
                 <ref name="types"/>
             </list>
         </attribute>
         <attribute name="class">
             <list>
-                <ref name="classes"/>
-                <value>notebody</value>
                 <ref name="classes"/>
             </list>
         </attribute>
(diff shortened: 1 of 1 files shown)


josteinaj:
I was not aware of the `<p class="verse-title">` markup. Wouldn't it make sense to mark it as a bridgehead instead? Or as an option? @AndersEkl @martinpub

AndersEkl:
> 
> 
> I was not aware of the `<p class="verse-title">` markup. Wouldn't it make sense to mark it as a bridgehead instead? Or as an option? @AndersEkl @martinpub

It does, yes. @martinpub , maybe we should change this right away? Using bridgehead.

kalaspuffar:
Hi @AndersEkl and @martinpub 

I've changed the implementation to use bridgehead. Easy to revert that change if needed.

Best regards
Daniel

josteinaj:
@martinpub @AndersEkl This is more of a guidelines question and not really a validator question: I'm curious about what the difference is between `line_indent` and `line_longindent`? Are there three types of indentation in poems (not indented, somewhat indented, severely indented)? It is not described in the guidelines as far as I can tell, but it might be more obvious to those ordering and producing the EPUBs than me.

src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -2088,55 +2087,66 @@
...
-            <!-- Strict content model: (h1 | span | p | section | img | figure)* -->
-            <ref name="attlist.poem"/> <!-- @epub:type, @class, @id, @title, @xml:space, @xml:lang, @lang, @dir -->
-            <zeroOrMore>
+        <element name="div">
+            <attribute name="class">
+                <value>verse</value>
+            </attribute>
+            <optional>
                 <choice>
                     <ref name="title"/> <!-- h1, h2, h3, h4, h5, h6 -->
```
josteinaj, line 2099: I don't think headers are allowed inside the `<div class="verse">`. Only bridgeheads (previously verse-title) are allowed here, while the headers should be placed outside the `div` as a direct child of the surrounding `<section>` element.

A sorrounding section element with a header is no different than any other types of chapters, so it doesn't need any explicit validation. So it should be enough to just remove the `<ref name="title"/>`. And remove the surrounding `<choice>` since there's only one choice remaining.

src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -2088,55 +2087,66 @@
...
-                    <ref name="line"/> <!-- p -->
-                    <ref name="pagebreak.block"/> <!-- div -->
-                    <ref name="img"/> <!-- img -->
-                    <ref name="figure.imggroup"/> <!-- figure -->
-                    <ref name="sidebar"/> <!-- aside, figure -->
+                    <ref name="bridgehead"/>
                 </choice>
-            </zeroOrMore>
+            </optional>
+            <ref name="poem.linegroup"/>
```
josteinaj, line 2103: This requires exactly one linegroup. This should be wrapped in `<oneOrMore>`.

src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -2088,55 +2087,66 @@
...
+            <optional>
+                <element name="p">
```
josteinaj, line 2105: Poems are required to have bridgheads (poem-title) before any linegroup, and verse-author after all the linegroups. This is probably how a poem is normally structured, but it is not described in the guidelines and I suppose there could be cases where the order could be different. For instance if you want to list the author right after the bridgehead/verse-title/header, if a verse (in the sense "linegroup") has a title (i.e. bridgehead) of its own, or is written by a different author.

@martinpub @AndersEkl : what do you think? Should there be a required order here?

If we allow zero or more bridgeheads and zero or more authors, we could do something like:

```xml
<element name="div">
    <attribute name="class">
        <value>verse</value>
    </attribute>
    <zeroOrMore>
        <choice>
            <ref name="bridgehead"/>
            <!-- verse-author here -->
        </choice>
    </zeroOrMore>
    <ref name="poem.linegroup"/>
    <zeroOrMore>
        <choice>
            <ref name="poem.linegroup"/>
            <ref name="bridgehead"/>
            <!-- verse-author here -->
        </choice>
    </zeroOrMore>
</element>
```

src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -2088,55 +2087,66 @@
...
-    <define name="attlist.poem" combine="interleave">
-        <attribute name="epub:type">
-            <list>
-                <ref name="types"/>
+    <define name="poem.linegroup">
+        <element name="p">
+            <attribute name="class">
+                <value>linegroup</value>
+            </attribute>
+            <oneOrMore>
```
josteinaj, line 2119: This allows a line to contain only a line number. It is also possible for some lines to have line numbers, and some lines to not have line numbers. It's a bit relaxed, but this is also how it was handled in DTBook where this structure stems from, and the new guidelines doesn't put any restrictions on it either.

@martinpub @AndersEkl does the following look like it should be allowed?

```html
<div class="poem">
    <p class="linegroup">
        <span class="linenum">13</span>
    </p>
    <p class="linegroup">
        <span class="linenum">1</span><span class="line">Lorem ipsum</span><br/>
        <span class="line">Lorem ipsum</span><br/>
        <span class="line">Lorem ipsum</span><br/>
        <span class="linenum">4</span>
        <span class="linenum">3</span><span class="line">Lorem ipsum</span><br/>
        <span class="linenum">3</span>
    </p>
</div>
```

src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -2088,55 +2087,66 @@
...
+    </define>
+
+    <define name="poem.line">
+        <element name="span">
+            <attribute name="class">
                 <choice>
-                    <!-- By mistake, z:3998:verse is allowed on the poem in guidelines version 2015-1, so for now we'll allow suppliers to either follow the guidlines (z3998:verse) or do it correctly (z3998:poem) -->
-                    <value>z3998:verse</value>
-                    <value>z3998:poem</value>
+                    <value>line</value>
```
josteinaj, line 2130: In cases where we have some required classes, should we also open up for all other types of classes as well? For instance, what if we want to add a class during our production workflow that will make this invalid:

```html
<span class="line highlight">…</span> <!-- (just an example, we don't use this class in our books) -->
```

On the other hand, if we allow "line + any other class", we would allow things like:

```
<span class="line verse verse-author">…</span>
```

…so we would need to allow only things that are not already defined to have a special meaning. I'm not sure if that's possible in RelaxNG. Maybe it would be possible to refer to the "classes" definition, and then into the `<data type="string"/>` at the end, add a list of exclusions, which would be the list of predefined classes. It might be possible to do `<data type="string"><except><value>line</value><value>linegroup</value><!-- etc. --></except></data>`, but I haven't tested.

You don't need to change it in this PR, but maybe we should go through all class attributes and add support for that later (@martinpub and @AndersEkl might have opinions here as well).

josteinaj (changes requested):
All in all good changes, but I have some comments.

kalaspuffar:
Hi @josteinaj 

I've done the simple changes that don't require input from @AndersEkl or @martinpub.

Best regards
Daniel

josteinaj (commented):
Looks good. Let's wait for feedback from @martinpub and @AndersEkl before merging.

AndersEkl:
> 
> 
> @martinpub @AndersEkl This is more of a guidelines question and not really a validator question: I'm curious about what the difference is between `line_indent` and `line_longindent`? Are there three types of indentation in poems (not indented, somewhat indented, severely indented)? It is not described in the guidelines as far as I can tell, but it might be more obvious to those ordering and producing the EPUBs than me.

Yes, the idea is to have a few options to represent the visual layout of poems in printed form. There are all sorts of variants with indentations of differing length and other fanciful ideas, but we thought that having at least one short and one long indentation would be a good compromise. We mentioned these classes in the guidelines but haven't gone into detail exactly how to use them. This should be specified in editing instructions.

src/main/resources/xml/schema/2020-1/nordic-html5.rng
AndersEkl: @josteinaj Sounds good to me!

src/main/resources/xml/schema/2020-1/nordic-html5.rng
AndersEkl: I think it is reasonable to require that a bridgehead, if there is one, is the first child of the `<div class="verse">` element. I am not sure it is necessary to have any rules concerning placement of the author? But it should not be inside a linegroup.

src/main/resources/xml/schema/2020-1/nordic-html5.rng
AndersEkl: I can definitely see cases where there are lines containing only a line number. That should be allowed. I am not sure it is necessary to put too much effort into building rules about this. Speaking from my own perspective, we rarely use line numbers at all. I am fine with having a rather relaxed validation here. I am not sure if @martinpub has a different opinion.

src/main/resources/xml/schema/2020-1/nordic-html5.rng
josteinaj: Ok. Should we wait for feedback from @martinpub if he has an opinion about the placement of the author?

AndersEkl: Maybe we shouldn't be too strict on these rules? What if you have a couple of poems in a row, with separate bridgeheads and separate authors? Should we then require a separate `<div>` for each poem in order to enforce rules about placement of bridgehead and author? Or should we allow a bit of flexibilty? It may be enough to just check that neither bridgehead or author are placed inside a linegroup? But maybe @martinpub should say what he thinks as well?

josteinaj: Could each verse potentially have a bridghead? Either if they actually have a title, or just to say "Verse 1", "Verse 2" etc. (if that's how it's done in the original or requested in editing instructions).

AndersEkl: Yeah, I don't see why not? And what would be the real gain from having very strict rules about where to place a bridgehead or an author name inside a verse? Is it necessary? We don't have any rules about placement of bridgeheads elsewhere, right?

josteinaj: I think the only rule about bridgeheads in other places is that they can only occur as a direct child of `<section>` (same as for normal headings), but I'm not sure.

kalaspuffar:
Hi @josteinaj 

Are you still waiting for an update from me? Read through this issue, and I don't think we have any outstanding questions? Or is there something I've missed?

Best regards
Daniel

josteinaj:
Hi @kalaspuffar. No I'm not waiting on you. We're waiting on feedback from @AndersEkl and @martinpub regarding:

- ordering of bridgehead(s) / linegroup(s) / author(s)
- presence and validation of linenum/line

AndersEkl:
I gave my thoughts above...
- I don't think it is important to have strict rules for the order of bridgehead(s), linegroup(s) and author(s).
- Line numbers should be a the start of a line, but otherwise I don't think we need many other rules. There can be cases where a line only has a number, and nothing else. There are also lots of cases where for instance every fifth line is numbered. So, I think we need a quite relaxed approach here.

josteinaj:
Ok.

So if we're quite relaxed about the ordering of bridgehead(s) / linegroup(s) / author(s), should we at least require a minimum of one linegroup? For instance: "*any number of bridgehead, linegroup and author in any order, but at least one linegroup*". My suggestion here should work I think: https://github.com/nlbdev/nordic-epub3-dtbook-migrator/pull/427#discussion_r581790075

For lines, we can be relaxed about that as well, in which case it's fine the way it is now. Just note that since these are inline elements, the CSS rules might become more complex if we have to account for linenums without a corresponding line element, unnumbered lines in the same linegroup as numbered lines, etc.

By the way, could it make sense to both "line" and "line_indent"/"line_longindent" together? Indentation styles could be attached to the line_indent and line_longindent classes, while CSS rules common to all lines, as well as cases where we want to search for or match lines regardless of their indentation, could use simply "line".

So for instance:

```html
<span class="line line_indent">…</span>
```

?

martinpub:
I think your suggestions seem reasonable @josteinaj

AndersEkl:
Yep, I agree!

kalaspuffar:
Hi @josteinaj 

Did minor changes to the poem rules. Now we allow the author or bridgehead before or after at least one line of the poem. The line contains tags with one or more of the classes line, line_indent or line_longindent.

Best regards
Daniel

src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -2095,22 +2095,29 @@
...
-                <element name="p">
-                    <attribute name="class">
-                        <value>verse-author</value>
-                    </attribute>
-                </element>
-            </optional>
+            <zeroOrMore>
+                <ref name="bridgehead"/>
+                <ref name="poem.author"/>
+            </zeroOrMore>
```
josteinaj, line 2108: I think `<ref name="poem.linegroup"/>` is needed inside this `<zeroOrMore>` to allow for for instance bridgeheads between linegroups.

src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -2126,11 +2133,11 @@
     <define name="poem.line">
         <element name="span">
             <attribute name="class">
-                <choice>
+                <oneOrMore>
```
josteinaj, line 2136: My thought here was to always require `line`, and make it optional to also have either `line_indent` or `line_longindent`. But this is probably fine as well, so that we're not too strict about it.

josteinaj (changes requested):

src/main/resources/xml/schema/2020-1/nordic-html5.rng
kalaspuffar: Great catch, had to think about it a bit but just adding it to the last clause would be ok.

josteinaj (approved):

PR #435, opened 2021-03-18
More fixes from examples.

kalaspuffar:
Hi @josteinaj 

This package of fixes contains:

* Allow role, aria-label and aria-labelledby
* remove DTBook legacy docauthor and doctitle "The validator still seems to expect a single-page structure. I think this is in the RelaxNG file." - @martinpub 
* revert version to 3.0. "This is incorrect, it should be 3.0 also for EPUB 3.2, see https://www.w3.org/publishing/epub3/epub-packages.html#sec-package-elem." - @martinpub 
* Allow text in page breaks

The above change where we remove the header block is what I interpreted @martinpub's comment above. I understand that we don't want this legacy in the code when it's not specified in the documentation but I'm not familiar with the "single-page structure" so I did not know exactly what was required by this comment.

Best regards
Daniel

### DIFF
--- a/src/main/resources/xml/schema/2020-1/nordic-html5.rng
+++ b/src/main/resources/xml/schema/2020-1/nordic-html5.rng
@@ -704,16 +704,14 @@
                 <group>
                     <!-- content document -->
                     <ref name="attrs"/> <!-- @id, @title, @xml:space, @xml:lang, @lang, @dir, @epub:type, @class -->
-                    <oneOrMore>
-                       <element name="section">
-                           <choice>
-                               <ref name="cover"/> <!-- @epub:type, @id, @title, @xml:space, @xml:lang, @lang, @dir, @class, section -->
-                               <ref name="frontmatter"/> <!-- @epub:type, @id, @title, @xml:space, @xml:lang, @lang, @dir, @class, div, h1, p, hr, ol, ul, dl, pre, blockquote, img, figure, section, span, table, address, aside, math, article -->
-                               <ref name="bodymatter"/> <!-- @epub:type, @id, @title, @xml:space, @xml:lang, @lang, @dir, @class, h1, p, hr, ol, ul, dl, pre, div, blockquote, img, figure, section, span, table, address, aside, math, article -->
-                               <ref name="backmatter"/> <!-- @epub:type, @id, @title, @xml:space, @xml:lang, @lang, @dir, @class, div, h1, p, hr, ol, ul, dl, pre, blockquote, img, figure, section, span, table, address, aside, math, article -->
-                           </choice>
-                       </element>
-                   </oneOrMore>
+                    <element name="section">
+                        <choice>
+                           <ref name="cover"/> <!-- @epub:type, @id, @title, @xml:space, @xml:lang, @lang, @dir, @class, section -->
+                           <ref name="frontmatter"/> <!-- @epub:type, @id, @title, @xml:space, @xml:lang, @lang, @dir, @class, div, h1, p, hr, ol, ul, dl, pre, blockquote, img, figure, section, span, table, address, aside, math, article -->
+                           <ref name="bodymatter"/> <!-- @epub:type, @id, @title, @xml:space, @xml:lang, @lang, @dir, @class, h1, p, hr, ol, ul, dl, pre, div, blockquote, img, figure, section, span, table, address, aside, math, article -->
+                           <ref name="backmatter"/> <!-- @epub:type, @id, @title, @xml:space, @xml:lang, @lang, @dir, @class, div, h1, p, hr, ol, ul, dl, pre, blockquote, img, figure, section, span, table, address, aside, math, article -->
+                        </choice>
+                    </element>
                 </group>
                 <group>
                     <!-- navigation document -->
--- a/src/main/resources/xml/schema/2020-1/nordic-html5.rng
+++ b/src/main/resources/xml/schema/2020-1/nordic-html5.rng
@@ -220,7 +220,27 @@
             </list>
         </attribute>
     </define>
-    
+
+    <define name="attr.role">
+        <attribute name="role">
+            <list>
+                <ref name="roles"/>
+            </list>
+        </attribute>
+    </define>
+
+    <define name="attr.aria.label">
+        <attribute name="aria-label">
+            <ref name="Text"/>
+        </attribute>
+    </define>
+
+    <define name="attr.aria.labelledby">
+        <attribute name="aria-labelledby">
+            <ref name="Text"/>
+        </attribute>
+    </define>
+
     <define name="coreattrs.base">
         <optional>
             <ref name="attr.id"/> <!-- @id -->
@@ -230,6 +250,15 @@
         </optional>
         <optional>
             <ref name="attr.xmlspace"/> <!-- @xml:space -->
+        </optional>
+        <optional>
+            <ref name="attr.role"/> <!-- @role -->
+        </optional>
+        <optional>
+            <ref name="attr.aria.label"/> <!-- @aria-label -->
+        </optional>
+        <optional>
+            <ref name="attr.aria.labelledby"/> <!-- @aria-labelledby -->
         </optional>
     </define>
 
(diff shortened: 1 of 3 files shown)


src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -231,6 +251,15 @@
...
         </optional>
+        <optional>
+            <ref name="attr.role"/> <!-- @role -->
+        </optional>
+        <optional>
+            <ref name="attr.aria.label"/> <!-- @aria-label -->
+        </optional>
+        <optional>
+            <ref name="attr.aria.labelledby"/> <!-- @aria-labelledby -->
+        </optional>
```
josteinaj, line 262: I think this will cause problems in instances where either role, aria-label or aria-labelledby are defined explicitly.

For instance, this is not valid RelaxNG:

```xml
<grammar xmlns="http://relaxng.org/ns/structure/1.0">
    <start>
        <element name="test">
            <attribute name="role"/>
            <optional>
                <attribute name="role"/>
            </optional>
        </element>
    </start>
</grammar>
```

It gives me the error `duplicate attribute "role"` in Oxygen XML Editor.

This is why we have the `*.notype` and `*.noclass` definitions, so that we can reference those when a `epub:type` or `class` have been explicitly defined.

See for instance:
https://github.com/nlbdev/nordic-epub3-dtbook-migrator/blob/45007dd2616e82e42028eb012e139879b8cd79e2/src/main/resources/xml/schema/2020-1/nordic-html5.rng#L743-L751

When both a `epub:type` and `class` is defined, we could use `attrsrqd.base` (this one) directly, as it essentially meant "core attributes except epub:type and class". Here's an example of this from the 2015-1 version of this file:
https://github.com/nlbdev/nordic-epub3-dtbook-migrator/blob/8f80992fbfcc24799ce99a071e41a2690611cfef/src/main/resources/xml/schema/nordic-html5.rng#L1633-L1648

So since we now have three new attributes, it gets more complicated. We might need a `coreatts.norole`, `coreatts.nolabel`, `coreatts.nolabelledby` for the cases where those attributes are not defined. Other combinations might even be useful as well, like `coreattrs.norolenoclass` if both a `role` and a `class` attribute are explicitly defined.

src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -346,6 +375,15 @@
...
```
josteinaj, line 386: Same issue here as for `coreattrs.base`.

src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -664,21 +702,8 @@
...
-                        <optional>
-                            <a:documentation>p</a:documentation>
-                            <ref name="covertitle"/> <!-- p -->
-                        </optional>
-                        <zeroOrMore>
-                            <a:documentation>p</a:documentation>
-                            <ref name="docauthor"/> <!-- p -->
-                        </zeroOrMore>
-                    </element>
                     <oneOrMore>
```
josteinaj, line 707: This `<oneOrMore>` should be removed as well. There will always be exactly one top-level section in an EPUB according to the 2020-1 guidelines.

src/main/resources/xml/schema/2020-1/nordic2020-1.sch
```diff
@@ -331,11 +331,6 @@
         <rule context="html:*[tokenize(@epub:type, '\s+') = 'pagebreak']">
             <let name="context" value="concat('(&lt;', name(), string-join(for $a in (@*) return concat(' ', $a/name(), '=&quot;', $a, '&quot;'), ''), '&gt;)')"/>
             <assert test="tokenize(@class, '\s+') = ('page-front', 'page-normal', 'page-special')">[nordic105] Page breaks must have either a 'page-front', a 'page-normal' or a 'page-special' class. <value-of select="$context"/></assert>
-            <assert test="count(* | comment()) = 0 and string-length(string-join(text(), '')) = 0">[nordic105] Pagebreaks must not contain anything<value-of select="
-                    if (string-length(text()) &gt; 0 and normalize-space(text()) = '') then
-                        ' (element contains empty spaces)'
-                    else
-                        ''"/>. <value-of select="$context"/></assert>
```
josteinaj, line 338: I think we want to keep a variant of this one:

```xml
            <assert test="count(* | comment()) = 0 and text() = normalize-space(text())">[nordic105] Pagebreaks must not contain elements or comments. Text content, if present, must not contain extra whitespace. <value-of select="$context"/></assert>
```

@martinpub @AndersEkl: right?

josteinaj (changes requested):

josteinaj:
The legacy single-page structure was:

```html
<html>
  <head>…</head>
  <body>
    <header><!-- doctitle / docauthor --></header>
    <section><!-- chapter 1 --></section>
    <section><!-- chapter 2 --></section>
  </body>
</html>
```

And the equivalent multipage structure was:

```html
<html>
  <head>…</head>
  <body><!-- chapter 1 --></body>
</html>
```

```html
<html>
  <head>…</head>
  <body><!-- chapter 2 --></body>
</html>
```

The `<header>` element mapped the `<doctitle>`, `<docauthor>` and `<covertitle>` elements from DTBook to HTML, but also functioned as a sort of "flag" for validation rules to see if it was a single-page HTML file or a multi-page HTML file. No `<header>` means multi-page. In the schematrons you probably noticed several matches for `body[header]` which is a way of matching only body elements in single-document versions.

src/main/resources/xml/schema/2020-1/nordic-html5.rng
kalaspuffar: Hi @josteinaj 

I see your point, and there might be a reason to specify a ```coreatts.norole```  if we want to create a specific role to decide a structure. We don't have that need yet, but we could make that change in the future.

When it comes to ```aria-label``` and ```aria-labelledby```  there is no such need IMHO.

The label can be any text and will not specify a type and the labeled by id is just a pointer.

Best regards
Daniel

kalaspuffar:
Hi @josteinaj 

I've implemented your suggestions. Please review again.

Best regards
Daniel

src/main/resources/xml/schema/2020-1/nordic-html5.rng
josteinaj: Good points.

Then maybe `aria-describedby` also should be declared as an attribute that can occur anywhere? If so, we wouldn't need to declare it in the image figure declaration (see discussion in #432).

kalaspuffar:
Hi @josteinaj 

Great suggestion, I've changed ```describedby``` to be a more general implementation. Please review again.

Best regards
Daniel

josteinaj (approved):

josteinaj:
The only thing remaining is https://github.com/nlbdev/nordic-epub3-dtbook-migrator/pull/435#discussion_r596825648

@martinpub @AndersEkl This rule was too strict:

- `[nordic105]`: *Pagebreaks must not contain anything*

So we could remove it completely, but I suggest that we instead replace it with:

- `[nordic105]`: *Pagebreaks must not contain elements or comments. Text content, if present, must not contain extra whitespace.*

Would you be fine with that?

josteinaj:
@kalaspuffar I approved this for merging a bit prematurely, sorry. I thought I only approved some of the changes and not the whole PR, but maybe that's not how reviews work. I should have mentioned the question about `nordic105` before approving.

But it's not a big deal. I see you included the change, which I think is good. If @martinpub or @AndersEkl wan'ts to remove it completely it could be removed in a later PR.

kalaspuffar:
Hi @josteinaj 

I've added that rule to the repository and I think it seems reasonable to not allow empty tags or just spaces. 

Best regards
Daniel

AndersEkl:
> 
> 
> The only thing remaining is [#435 (comment)](https://github.com/nlbdev/nordic-epub3-dtbook-migrator/pull/435#discussion_r596825648)
> 
> @martinpub @AndersEkl This rule was too strict:
> 
>     * `[nordic105]`: _Pagebreaks must not contain anything_
> 
> 
> So we could remove it completely, but I suggest that we instead replace it with:
> 
>     * `[nordic105]`: _Pagebreaks must not contain elements or comments. Text content, if present, must not contain extra whitespace._
> 
> 
> Would you be fine with that?

Seems reasonable to me, yes.

martinpub:
> Seems reasonable to me, yes.

Agreed!

PR #437, opened 2021-03-24
Allow EPUBs without nav.ncx

kalaspuffar:
Hi @josteinaj 

Here is the result of our discussion today.

Best regards
Daniel

### DIFF
--- a/src/main/resources/xml/xproc/step/2020-1/epub3-validate.step.xpl
+++ b/src/main/resources/xml/xproc/step/2020-1/epub3-validate.step.xpl
@@ -106,7 +106,7 @@
 
 
             <p:variable name="basedir" select="if (/*/d:file[@media-type='application/epub+zip']) then $temp-dir else base-uri(/*)"/>
-            
+
             <p:choose>
                 <p:when test="$use-epubcheck = 'true'">
                     <p:choose name="epub3-validate.step.choose-zipped-or-not">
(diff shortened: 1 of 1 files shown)


src/main/resources/xml/xproc/step/2020-1/epub3-validate.step.xpl
```diff
@@ -253,7 +253,7 @@
                     <p:pipe port="in-memory" step="epub3-validate.step.unzip"/>
                 </p:input>
             </px:fileset-load>
-            <px:assert test-count-min="1" test-count-max="1" message="There is no NCX with the filename 'nav.ncx' in the EPUB" error-code="NORDICDTBOOKEPUB014"/>
+            <px:assert test-count-min="0" test-count-max="1" message="There is no NCX with the filename 'nav.ncx' in the EPUB" error-code="NORDICDTBOOKEPUB014"/>
```
josteinaj, line 256: The error text is a bit misleading now that the NCX is optional, and I think the test will in practice always succeed anyway, because of the `<p:delete match="/*/*[not(ends-with(@href,'nav.ncx'))]"`. I guess it would catch cases where there are multiple files called `nav.ncx` but in different directories. Validation of whether or not there are multiple NCX files, and whether or not they have the correct file name is something that is better to do in the opf schematron validation anyway, so I think this is fine as it is.

josteinaj (approved):

josteinaj:
I haven't tested it, but the code looks good, so I'll merge.

Issue #444, opened 2021-03-30
Validator issue: [nordic_nav_references_2b] The nesting of headlines in the content does not match the nesting of headlines in the navigation document


martinpub:
Validator result for DTB37628:

`[nordic_nav_references_2b] The nesting of headlines in the content does not match the nesting of headlines in the navigation document. The toc item (<a href="DTB37628-001-cover.xhtml#prodnote_2">) in the navigation document is not nested at the correct level. The referenced document (DTB37628-001-cover.xhtml) occurs in the navigation document at nesting depth 1 (it is not contained inside other sections such as a part or a chapter). The referenced headline (DTB37628-001-cover.xhtml#prodnote_2) occurs in the navigation document at nesting depth 2 (it is contained inside "Cover"). The referenced headline (<></>) occurs in the content document DTB37628-001-cover.xhtml as a  which implies that it should be referenced at nesting depth 3 in the navigation document.`

However, looking at nav.xhtml, it seems to match the two-level section nesting (correctly) present in the content document:

```
<nav role="doc-toc" aria-labelledby="n1" epub:type="toc">
<h1 id="n1">Contents</h1>
<ol class="plain">
<li><a href="DTB37628-001-cover.xhtml#level1_1">Omslag</a>
<ol class="plain">
<li><a href="DTB37628-001-cover.xhtml#prodnote_1">Framsida</a></li>
<li><a href="DTB37628-001-cover.xhtml#prodnote_2">Baksida</a></li>
</ol>

```

This is the start of the content document:

```
<body>
<section id="level1_1" epub:type="cover" aria-label="Omslag">
<section class="frontcover" id="prodnote_1" aria-label="Framsida">
<figure class="image"><img alt="image" src="images/cover.jpg"/>
</figure>
</section>
<section class="rearcover" id="prodnote_2" aria-label="Baksida">
```

Issue #443, opened 2021-03-30
Validator issue: [nordic_nav_references_1] All headings in the content documents must be referenced exactly once in the navigation document.


martinpub:
> `[nordic_nav_references_1] All headings in the content documents must be referenced exactly once in the navigation document. In the content document "DTB37628-009-chapter.xhtml", the heading "2. Kuppartad inledning – i god demokratisk ordning? 1994–2000" with id="h1_6" inside the "section" element is not referenced from the navigation document.`

(from DTB37628-corrected validation)

This does not seem correct. 

The content doc looks like this:

```
<section id="level1_9" role="doc-chapter" aria-labelledby="h1_6" epub:type="bodymatter chapter">
<div epub:type="pagebreak" role="doc-pagebreak" class="page-normal" id="page-33" aria-label="33"/>
<h1 id="h1_6">2. Kuppartad inledning &#x2013; i god demokratisk ordning? 1994&#x2013;2000</h1>
```

And the section "level1_9" is correctly referenced in nav.xhtml. Having the headings referenced directly is something we actively want to avoid, so the check should be to check that all sections are covered by nav.xhtml?

Issue #442, opened 2021-03-30
Validator issue: [xhtml] Line: 13 Col: 1705 text not allowed here; expected the element end-tag  (X60352A-13-chapter.xhtml)


martinpub:
Markup triggering this error:

`<li id="li_9" value="9">Rist'kateisei-s'je t&#xf6; muga: Tat meiden, kudam taivhil! i sv&#xe4;tikahase nimi sinun.<span id="Page_24" class="page-normal" epub:type="pagebreak" role="doc-pagebreak" title="24">24</span></li>`

Is this an EPUBCheck error? It seems to be a correct use of span inside an li?

Ping @kalaspuffar @AndersEkl

kalaspuffar:
Hi @martinpub 

There is a couple of different rules in the validation report.
```
[epubcheck]
[ace]
[xhtml] = RelaxNG
[nordic_*] = Schematron
```

Sadly the RelaxNG rules are the most in need of review, and there is a lot to change. In this case, I forgot to add Text as an allowed child of a page break span. We allowed it for block page breaks.

Best regards.
Daniel

martinpub:
I see. Thanks for explaining the different rule names!

Issue #441, opened 2021-03-30
Validator issue: [nordic_nav_ncx_5] Landmarks will result in duplicate references, so the rule should be loosened up


martinpub:
Example:

> `[nordic_nav_ncx_5] Two references in the navigation document can not point to the same location in the content. (<a epub:type="colophon" href="X60352A-03-colophon.xhtml#f2">)`

Landmarks addition to guidelines will trigger duplicate references in the nav doc. restrict rule to require unique references only within each `<nav>` block?

`@class="success"` seems like a bug?

Ping @kalaspuffar @AndersEkl

kalaspuffar:
Hi @martinpub 

All the report rules are marked green and are information to the operator but not a direct violation of the specification.

```
<report test="@href = preceding::html:a/@href">[nordic_nav_ncx_5] Two references in the navigation document can not point to the same location in the content. <value-of select="$context"/></report>
```

Assertions, on the other hand, are incorrect behavior that is not allowed by the specification e.g.
```
<assert test="self::html:li">[nordic203d] 'endnote' can only be applied to &lt;li&gt; elements. <value-of select="$context"/></assert>
```

If you don't want this distinction, we can handle all information as erroneous.

Best regards
Daniel

martinpub:
Thanks for the explanation.

I would say then that report rules should result in a "warning", perhaps warning is a more appropriate HTML class, and perhaps green is not the best color. However, then the report text documentation should also be phrased as such, i.e. something like "Two references in the navigation document  currently point to the same location in the content."

This was a more general comment, I think in this particular case, we probably want to change the warning to be restricted as suggested above, otherwise we will have a lot of warnings ...

josteinaj:
I may have used `<report>` the wrong way last time the schematrons were revised. I have been thinking of it as simply the inverse of `<assert>`, so that `<report test="…">` would be the same as `<assert test="not(…)">`, but a quick search indicates that this is not quite the case.

PR #445, opened 2021-03-31
Fixes after review of reported issues.

kalaspuffar:
Hi @josteinaj 

* List can have page breaks in structure. 
* Changed references from warning to an error. #441
* Inline page breaks can have text. #442
* Updated reference transform script removing body element and use role instead of type. (list-heading-and-pagebreak-references.xsl) #443, #444
* Added fix so we only fetch the title span when it's available and compare that with the toc. #446

Best regards
Daniel


src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -1571,6 +1571,7 @@
                     <ref name="inlinenoprodnote"/> <!-- em, strong, dfn, code, samp, kbd, span, abbr, math, a, img, br, q, sub, sup, bdo -->
                     <ref name="lic"/> <!-- span -->
                     <ref name="list"/> <!-- ol, ul -->
+                    <ref name="pagebreak.block"/> <!-- div -->
```
josteinaj, line 1574: I think this should be `pagebreak.inline`. The `<choice>` here chooses between inline content (the first `<choice>`) and block content (the second `<choice>`). The reference to `list` here allows for how navigation documents are marked up.

josteinaj (changes requested):
I'm not sure if this addresses #444 or not. Maybe the modified `list-heading-and-pagebreak-references.xsl` file does, but the error message should maybe be updated anyway?

src/main/resources/xml/schema/2020-1/nordic-html5.rng
kalaspuffar: Hi @josteinaj 

This will solve the issue in question; a block page break was used inside of a <LI> before the definition of another <OL> element. The inline page break is used mostly inside of paragraphs. Perhaps it's allowed here inside the inlinenoprodnote rule.

Best regards
Daniel

kalaspuffar: Perhaps @martinpub or @AndersEkl could shed some light here. Do we allow block page breaks inside of list items?

AndersEkl: It depends on the context. If the list item only has text inside it, with no p or other block elements, only inline page breaks would be allowed.

But if the list items contains other block elements, such as paragraphs or nested lists, then a block page break between other block elements would be ok, I'd say.

kalaspuffar: Hi @josteinaj 

Did this issue resolve itself, or do you still want me to change something?

Best regards
Daniel

kalaspuffar:
Hi @josteinaj 

Added a fix for #446 as suggested by @martinpub 

This will fetch the title span if it's available and if not use all of the title element to compare with the toc link.

Please review again.

Best regards
Daniel

src/main/resources/xml/schema/2020-1/nordic-html5.rng
kalaspuffar: I've now changed this to inline.

PR #451, opened 2021-05-03
Removing either.

kalaspuffar:
Hi @martinpub 

This test checks that each sectioning element has a reference, not that it's unique. The next rule does that testing. So I thought it was reasonable to only remove the word either from the message.

Best regards
Daniel


martinpub:
Great, thanks @kalaspuffar!

Issue #449, opened 2021-04-07
Validator issue: linegroups are not allowed in non-verse settings


martinpub:
The testbook DTB38216 had an example of linegroup in a non-verse setting, but it seems the RelaxNG schema forbids it?

These constructions

```
<section aria-labelledby="hd9-2" id="level9-2"><h2 id="hd9-2">MEAN GIRLS? GENDER PORTRAYALS IN TEEN MOVIES</h2>
<p><strong>Elizabeth Behm&#x002D;Morawitz and Dana E. Mastro</strong></p>
<p class="linegroup"><span class="line"><span class="linenum">1</span> Research examining the effects of media exposure demonstrates that media consumption has a measurable</span>
```

triggered these errors in the validator:

> [xhtml] (Line: 00029 Column: 00022) value of attribute "class" is invalid

I don't remember this being a restriction in the previous guidelines, and I think semantically it would be good to be able to mark even non-verse content with linenumber/line markup. (This case was an appendix to a methodology book where an article is printed verbatim with numbered lines for simplyfing referencing from inside the book.)

Ping @AndersEkl @kalaspuffar

kalaspuffar:
Hi @martinpub 

I've looked at the specification, and the only place that we mention 'linenum' is in '3.4.11 Poetry and Verse'.

Changing this, so we allow line groups everywhere is possible but not recommended. Allowing specific content in a context where it isn't easily handled would add more complexity for the Reading Systems developers.

The point of a specification is to be concise to enable reading. Allow line groups everywhere could create a complexity of books with a less desirable layout. 

I remember a dtbook we analyzed where each line of text in the book had a line group.

Best regards
Daniel

martinpub:
Thanks for checking @kalaspuffar. I agree that this is an item for clarification in the guidelines. I would argue the use case is there to allow it in other settings, but perhaps it then needs a separate wrapper on the same level as the poetry/verse wrapper. I will create an issue for a second version of the guidelines.

martinpub:
See also https://github.com/nlbdev/epub3-guidelines-update/issues/137.

martinpub:
This is still an issue for MTM, as we have content we need to mark up line by line that is not verse/poems. We would like to have the validator accept the markup outside of verse settings until the guidelines have been updated to clarify this use case.

martinpub:
[...]

PR #454, opened 2021-05-06
Updated handling of line-by-line content.

kalaspuffar:
Hi @martinpub 

This is a solution to handle line-by-line content allowing construction with the linenum both inside of the line and outside in the linegroup.

Example of allow constructions:
```
<div class="verse">
    <p class="linegroup"><span class="line"><span class="linenum">1</span> This is line one</span></p>
    <p class="linegroup"><span class="line"><span class="linenum">2</span> This is line two</span></p>
    <p class="linegroup"><span class="line"><span class="linenum">3</span> This is line three</span></p>
    <p class="linegroup"><span class="line"><span class="linenum">4</span> This is line four</span></p>
</div>

<div class="verse">
    <p class="linegroup"><span class="linenum">1</span><span class="line"> This is line one</span></p>
    <p class="linegroup"><span class="linenum">2</span><span class="line"> This is line two</span></p>
    <p class="linegroup"><span class="linenum">3</span><span class="line"> This is line three</span></p>
    <p class="linegroup"><span class="linenum">4</span><span class="line"> This is line four</span></p>
</div>

<div class="line-by-line">
    <p class="linegroup"><span class="line"><span class="linenum">1</span> This is line one</span></p>
    <p class="linegroup"><span class="line"><span class="linenum">2</span> This is line two</span></p>
    <p class="linegroup"><span class="line"><span class="linenum">3</span> This is line three</span></p>
    <p class="linegroup"><span class="line"><span class="linenum">4</span> This is line four</span></p>
</div>

<div class="line-by-line">
    <p class="linegroup"><span class="linenum">1</span><span class="line"> This is line one</span></p>
    <p class="linegroup"><span class="linenum">2</span><span class="line"> This is line two</span></p>
    <p class="linegroup"><span class="linenum">3</span><span class="line"> This is line three</span></p>
    <p class="linegroup"><span class="linenum">4</span><span class="line"> This is line four</span></p>
</div>
```

Trying to solve
https://github.com/nlbdev/nordic-epub3-dtbook-migrator/issues/449

### DIFF
--- a/src/main/resources/xml/schema/2020-1/nordic-html5.rng
+++ b/src/main/resources/xml/schema/2020-1/nordic-html5.rng
@@ -2143,6 +2143,7 @@
     <define name="poem">
         <a:documentation>  Use: poem is a complete poem or fragment thereof.  </a:documentation>
         <element name="div">
+            <ref name="attrs.base"/>
             <attribute name="class">
                 <choice>
                     <value>verse</value>
--- a/src/main/resources/xml/schema/2020-1/nordic-html5.rng
+++ b/src/main/resources/xml/schema/2020-1/nordic-html5.rng
@@ -2144,19 +2144,26 @@
         <a:documentation>  Use: poem is a complete poem or fragment thereof.  </a:documentation>
         <element name="div">
             <attribute name="class">
-                <value>verse</value>
+                <choice>
+                    <value>verse</value>
+                    <value>line-by-line</value>
+                </choice>
             </attribute>
             <zeroOrMore>
-                <ref name="bridgehead"/>
-                <ref name="poem.author"/>
+                <choice>
+                    <ref name="bridgehead"/>
+                    <ref name="poem.author"/>
+                </choice>
             </zeroOrMore>
             <oneOrMore>
                 <ref name="poem.linegroup"/>
             </oneOrMore>
             <zeroOrMore>
-                <ref name="poem.linegroup"/>
-                <ref name="bridgehead"/>
-                <ref name="poem.author"/>
+                <choice>
+                    <ref name="poem.linegroup"/>
+                    <ref name="bridgehead"/>
+                    <ref name="poem.author"/>
+                </choice>
             </zeroOrMore>
         </element>
     </define>
@@ -2198,7 +2205,13 @@
                 </list>
             </attribute>
             <zeroOrMore>
+                <ref name="poem.linenum"/>
+            </zeroOrMore>
+            <zeroOrMore>
                 <ref name="inline"/> <!-- em, strong, dfn, code, samp, kbd, span, abbr, math, a, img, br, q, sub, sup, bdo -->
+            </zeroOrMore>
+            <zeroOrMore>
+                <ref name="poem.linenum"/>
             </zeroOrMore>
         </element>
         <optional>
@@ -2211,6 +2224,7 @@
             <attribute name="class">
                 <value>linenum</value>
             </attribute>
+            <ref name="Text"/>
         </element>
     </define>
 
--- a/src/main/resources/xml/schema/2020-1/nordic-html5.rng
+++ b/src/main/resources/xml/schema/2020-1/nordic-html5.rng
@@ -2173,6 +2173,7 @@
             <attribute name="class">
                 <value>verse-author</value>
             </attribute>
+            <ref name="Text"/>
         </element>
     </define>
 
--- a/src/main/resources/xml/schema/2020-1/nordic-html5.rng
+++ b/src/main/resources/xml/schema/2020-1/nordic-html5.rng
@@ -2187,7 +2187,6 @@
             <oneOrMore>
                 <choice>
                     <ref name="poem.line"/>
-                    <ref name="poem.linenum"/>
                 </choice>
             </oneOrMore>
         </element>


martinpub:
Thanks @kalaspuffar,

I think it looks good, but I think we should always require that `<span class="linenum">` be children to `<span class="line">`. What do you think @AndersEkl? That makes more sense semantically I think, that the line number is part of the line contents, however separated from other contents. So I would say in your examples, @kalaspuffar, number 2 and 4 should not be considered valid.

AndersEkl:
@martinpub I see your point, but having them in separate `<span>` elements next to each other also has some advantages. I think you'd have more flexibilty with styling options. I don't know if we can say there are any semantic considerations here either way, as `<span>` has no semantic role in itself and we only add class attributes. Are there any obvious problems with accepting both options?

kalaspuffar:
Hi @AndersEkl and @martinpub 

If we aren't required to support both then parsing and handling of the data will be easier for consumers so if we don't see a real benefit of having both options then I'm all for removing one of them.

Best regards
Daniel

AndersEkl:
@kalaspuffar I see...
So, we could see `<span class="line">` either as:
- the whole line incl line number, in which case `<span class="linenum">` must be a child
- always just the line of text, in which case `<span class="linenum">` should be a sibling

How were line numbers handled in DTBook? I don't remember, as we always stripped them away.

martinpub:
> How were line numbers handled in DTBook? I don't remember, as we always stripped them away.

Hmm ... I think my gut feeling is grounded in DTBook semantics, but I'm not sure. Will check it and get back here.

martinpub:
2015-1 guidelines:

> 3.1.4.18 Linegroup: `<div class="linegroup">`
The `<div class="linegroup">` tag is used to preserve the formatting of text grouped into line sets. The `<p class="line">` tag is used to wrap the individual lines within the linegroup.
The inclusion of a `<span>` element is required when line numbering is present. The attribute `class="linenum"` must be applied to the `<span>` element containing the number content. One example of books requiring this type of markup is language textbooks.

AndersEkl:
Ah, right... `<p>` was used for each line. That means you had to put the line number inside the `<p>` element. I always had an issue with `<p>` being used for something that often wasn't an actual paragraph. I think that is semantically incorrect. But, that is not the issue here... I don't know what I think here. Like I said, I see some advantages with having line numbers separate from the line itself. But I can't really say that one thing is correct and the other is incorrect.

martinpub:
From DTBook DTD (2005-3):

> `<!ELEMENT line (%inline; | linenum)* >`

I suggest then to stick to the DTBook-derived semantics, which I think is implied in the 2015-1 Nordic guidelines. Less confusion for colleagues as well as for external producers.

I think CSS is flexible enough to give good opportunities anyway. Or do you have a specific situation in mind @AndersEkl?

AndersEkl:
It would still be inline in either options. But I guess it would be convenient to have something that resembles the DTBook markup. I'm fine with either, to be honest. I was just trying to see if there were anything we should consider here.

I was thinking of maybe having a styling so that the actual text lines have a straight left margin and the numbers sort of floating outside of it. Maybe you could still achieve that? Not sure.

kalaspuffar:
Hi @AndersEkl and @martinpub 

IMHO we should keep semantics and styling separate. Most things are achievable in css without changing mark-up. Only exception might be when you have two few elements.

In the specification the semantics of what we want to express is the important part and the styling is a separate issue.

Best regards
Daniel

martinpub:
> I was thinking of maybe having a styling so that the actual text lines have a straight left margin and the numbers sort of floating outside of it. Maybe you could still achieve that? Not sure.

```
.linenum {
  float: left;
}
```
or something like that? I think a CSS ninja (that is, not me) could fix it quite elegantly.

AndersEkl:
@kalaspuffar
Yes, I would never let any styling considerations mess up semantics. But in this case the semantics are pretty much unspecified. But, let's go with Martin's gut feeling here. :)

egli:
I'd also prefer if we stick to DTBook-derived semantics, i.e. `linenum` as part of `line`. The [DAISY 3 Structure Guidelines](http://www.daisy.org/z3986/structure/SG-DAISY3/index.html) have a section about [Line Numbers](http://www.daisy.org/z3986/structure/SG-DAISY3/part2-inline.html#linenum) where they give the following example:
``` xml
<p>
  <line><linenum>1</linenum> There was a little girl</line>
  <line><linenum>2</linenum> Who had a little curl</line>
</p>
```
So, I think we should only allow example 1 and 3 above.

kalaspuffar:
Small update so now we only support these two cases.

```
<div class="verse">
  <p class="linegroup"><span class="line"><span class="linenum">1</span> This is line one</span></p>
  <p class="linegroup"><span class="line"><span class="linenum">2</span> This is line two</span></p>
  <p class="linegroup"><span class="line"><span class="linenum">3</span> This is line three</span></p>
  <p class="linegroup"><span class="line"><span class="linenum">4</span> This is line four</span></p>
</div>

<div class="line-by-line">
  <p class="linegroup"><span class="line"><span class="linenum">1</span> This is line one</span></p>
  <p class="linegroup"><span class="line"><span class="linenum">2</span> This is line two</span></p>
  <p class="linegroup"><span class="line"><span class="linenum">3</span> This is line three</span></p>
  <p class="linegroup"><span class="line"><span class="linenum">4</span> This is line four</span></p>
</div>
```

martinpub:
Thanks for your input everyone! Will merge this now.

Issue #463, opened 2021-05-25
Include aria-label-type section names in cross-nav-contentdoc check


martinpub:
Currently, rule [nordic_nav_references_1] checks the integrity of heading content so that toc entries in the navigation document should match the heading of the referred section in the content document.

The same integrity check should be done when a section does not have a h[x] heading tied to it, but only an aria-label attribute.

To take an example, the check should be performed to avoid mismatches such as:

Navigation toc entry: `<li><a href="T-000011-001-cover.xhtml#prodnote_1">Framsida</a></li>`
Content document: `<section aria-label="Front cover" class="frontcover" id="prodnote_1">`

They should match the same way as ordinary headings should, so the rule should be updated to include aria-label matching in the absence of a h[x] check.

Ping @kalaspuffar.

martinpub:
(Tested and verified.)

PR #467, opened 2021-05-26
Require headings to be equal to toc. 

kalaspuffar:
Hi @martinpub 

Require headings to be equal to toc. 
If there is no H[x] there should be an aria-label with the correct information.

This should solve #463.

Best regards
Daniel


martinpub:
Thanks @kalaspuffar!

Issue #460, opened 2021-05-21
Bridgehead is not allowed in `<aside>`


martinpub:
Current rule: [nordic270] Bridgehead is only allowed as a child of section, article and div. (`<p epub:type="bridgehead">`).

According to the guidelines text, they does not seem to be forbidden in asides:

> Headings in text boxes is by default required to be marked up using <p epub:type="bridgehead">. These are structurally unimportant headings and will not be included in the navigation document. If, however, certain text boxes should be included in the navigation document, they must be marked up using a <h[x]> element of an appropriate level, following the level structure of the surrounding text. Only text boxes marked up with <aside> can have <h[x]> headings. The Ordering Agency will specify in the Editing Instructions if <h[x]> markup is required.

Even though I know ACE complains when we have untitled (as in no aria-labelledby/aria-label) asides, but according to how the guidelines text is currently formulated, I think bridgehead should be allowed in asides, right @AndersEkl?

Ping @kalaspuffar

AndersEkl:
Yeah, that sounds reasonable to me.

kalaspuffar:
Hi @martinpub 

Bridgehead is allowed using ```epub:type```, but the example above is not a ```sidebar```. It's a ```text-box```. Sidebars need to have the class="sidebar" to distinguish from other aside elements.

I looked into https://format.mtm.se/nordic_epub/2020-1#sidebars-text-boxes-etc.

Does this mean that class ```sidebar``` or ```text-box``` should be interpreted as sidebars or is any other classes that will be used here?

Best regards
Daniel

Best regards
Daniel

martinpub:
Hi @kalaspuffar. For all purposes, the class "sidebar" was deprecated and replaced by class "text-box" with the 2020-1 guidelines. I hope this clarifies the issue.

Ping @AndersEkl

AndersEkl:
The class attribute "sidebar" was present to map elements to the DTBook element `<sidebar>`. The way `<sidebar>` was used previously was not semantically correct. It was used for all sorts of purposes, regardless of if the content was part of the main text or not. Semantically, `<sidebar>` should be only content that is not part of the main text. This function is very similar to the `<aside>` element in HTML. The class attribute "text-box" is only meant as an styling indicator and can be used for both secondary content, marked up with `<aside>`, and content that is integral to the main text, marked up with `<div>`. I don't see any need for the class attribute "sidebar" any more, as the semantic function is carried by which HTML element is used.

[...]

PR #469, opened 2021-06-03
Allow bridgehead in text-box

kalaspuffar:
These changes will solve issue #460


martinpub (changes requested):
Hi @kalaspuffar, I think the change of class "sidebar" to "text-box" is correct, however, "text-box" should not be a valid epub:type value, as it's not listed in https://idpf.github.io/epub-vocabs/structure/.

martinpub (approved):

PR #480, opened 2021-06-28
Updated tests to use the new guideline book which should validate.

kalaspuffar:
Hi @martinpub 

I've updated the 2020-1 tests so they will now validate against X60352A instead of C00000 and we now have passing tests for this repository.

Best regards
Daniel


martinpub (approved):
Thanks @kalaspuffar! I guess we will target remaining validation issues with the new test book in a separate issue.

PR #481, opened 2021-06-28
Fixing ASCIIMath markup in test book X50525A

martinpub:


martinpub:
Still merge problems ...

PR #484, opened 2021-06-29
Switch to newer epubcheck and modules parent.

kalaspuffar:
Hi @martinpub 

This will solve the issue we had with not creating a correct XML document. This was because we have updated to the latest EPUBCheck in the pipeline. 

I've updated to the latest EPUBCheck and the next to the latest version of the parent modules. Lastly, I've updated some namespaces to their new domains.

We are not changing to using the last version yet of pipeline modules because it will break the tests. The last version of the pipeline will not create valid EPUBs following the nordic guidelines. This could be good to keep in mind if you want to convert DTBooks to EPUB.

Best regards
Daniel

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
     <parent>
         <groupId>org.daisy.pipeline.modules</groupId>
         <artifactId>modules-parent</artifactId>
-        <version>1.13.7</version>
+        <version>1.14.1</version>
         <!--<relativePath>../../../parent/</relativePath>-->
     </parent>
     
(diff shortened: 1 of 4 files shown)


martinpub (approved):
Hi @kalaspuffar, and thank you very much for this fix. Does this mean that with this, the validator could be used together with the present state of the MTM Pipeline 2 edition?

kalaspuffar:
Hi @martinpub 

Yes, it can be used, but the EPUBs will not be valid if generated in the pipeline.

I have talked with @bertfrees about the issue and created an issue to describe the problem https://github.com/daisy/pipeline/issues/612

Best regards
Daniel

PR #485, opened 2021-06-30
Disable the pub identifier requirement.

kalaspuffar:
Hi @martinpub 

I've commented out the requirement to have unique-identifier set to pub-identifier.

Best regards
Daniel


martinpub:
Thanks @kalaspuffar!

For information, the reason that this is removed is that currently, Pipeline 2 EPUB generation does not preserve the name of the connecting attribute "pub-identifier", while still preserving integrity between dc:identifier and unique-identifier. A relaxation has been suggested in https://github.com/daisy/pipeline/issues/612.

It should be noted that already EPUBCheck verifies the presence of unique-identifier on the root element of package.opf, and that it is pointing to a corresponding meta tag with an ID.

We can get back to discussing this, after summer, but this will solve a production block at MTM and given the discussion above, the removal is not considered dangerous.

PR #486, opened 2021-06-30
Added logic to validate parts and chapters in different document

kalaspuffar:
Hi @martinpub 

This PR will handle validation of the correct structure when you separate parts and chapters in different files. I've also added more parts to the test case in order to test some edge cases.

Best regards
Daniel


martinpub (approved):
Thanks @kalaspuffar

Issue #487, opened 2021-07-02
Allow doc-acknowledgements in backmatter


martinpub:
Currently, this is forbidden by [nordic291].

Error:

> Backmatter can only use roles (doc-afterword, doc-appendix, doc-bibliography, doc-colophon, doc-conclusion, doc-dedication, doc-epigraph, doc-epilogue, doc-glossary, doc-index) (`<section id="level1_60" role="doc-acknowledgments" aria-labelledby="h1_56" epub:type="backmatter acknowledgments">`)

@kalaspuffar, could you have a look?

martinpub:
Verified fixed with #489.

PR #489, opened 2021-07-05
Adding the section type acknowledgments.

kalaspuffar:
Hi @martinpub 

This PR aims to solve the issue #487.

As a side note it would improve the process if you are able to create a commit / PR with an updated test book. Half the battle is to figure out how to add these constructs in the correct manor. And in some cases the specification is not clear either. In this case I was wondering if the acknowledgment could be a part of a backmatter file or if it needed to be a separate file. I opted to creating a new file.

As you are the expert on the format you could easily change the book to another valid state and then I change the validator so it will accept that state. Could also be good as a discussion point if we are unsure about the correct markup.

Just a thought.

Best regards
Daniel


martinpub:
Thank you very much @kalaspuffar!

And of course, adding to the test book is of course something I could provide you with.

martinpub (approved):
Hi @kalaspuffar, and thanks for this! I added a PR in #490 before having seen that you also provided a content file to the test book. Will overwrite this edition in merging that.

Issue #476, opened 2021-06-23
Decide on line break convention to use for this repo and harmonise content


martinpub:
From the discussion on today's validation group meeting.

Subtasks:

1. Check which type(s) is/are used
2. Decide on preferred one
3. Harmonise all content in one go

kalaspuffar:
I've done some initial tests and the checkstyle plugin might not be the best fit.

https://github.com/checkstyle/checkstyle/issues/614#issuecomment-73045960

"Checkstyle have small team and a huge plans on extending for java language, and it should focus on one language, to be at the top among java analysers."

PR #507, opened 2021-11-01
Add an editor config, a mvn goal and a ci action to check code conventions

egli:
This PR adds a cross editor config file, a Maven goal that uses the config to check the source code and finally a github action to check every commit for these conventions via maven.

Should fix at least in part #476.

When running the mvn goal quite a few files have problems. These can easily be fixed with `mvn editorconfig:format`. But I think such a global change should go in a separate PR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -179,7 +179,7 @@
                 <executions>
                     <execution>
                         <id>check</id>
-                        <phase>verify</phase>
+                        <phase>compile</phase>
                         <goals>
                             <goal>check</goal>
                         </goals>
--- a/pom.xml
+++ b/pom.xml
@@ -172,6 +172,28 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.ec4j.maven</groupId>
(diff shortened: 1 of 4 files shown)


egli:
Just FYI to see what the kind of code convention problems are check the [output of the github action](https://github.com/nlbdev/nordic-epub3-dtbook-migrator/runs/4065764715?check_suite_focus=true).

pom.xml
```diff
@@ -172,6 +172,28 @@
...
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.ec4j.maven</groupId>
+                <artifactId>editorconfig-maven-plugin</artifactId>
+                <version>0.1.1</version>
+                <executions>
+                    <execution>
+                        <id>check</id>
+                        <phase>verify</phase>
```
josteinaj, line 182: I'll move this to the `compile` phase if you don't mind. That way, if you run `mvn test`, editorconfig will be run before all the tests.

josteinaj:
Looks great, thanks @egli!

josteinaj (approved):

PR #514, opened 2022-05-10
Updating Docker commands

martinpub:
Hi @josteinaj, I updated to the Docker commands last known to work for me. Do you think the parameters are OK? I think the linking method was needed for proper exposure of web service on a Windows host. This is also one of the examples at the [DAISY Pipeline 2 web page](https://daisy.github.io/pipeline/Get-Help/User-Guide/Installation/).


martinpub:
Hi @josteinaj, is this change OK according to you?

josteinaj (commented):
Works for me

Issue #521, opened 2022-09-02
hr separator not allowed between code blocks and in poems


martinpub:
`<pre>` is a block-level element, but the following is not allowed according to the validator:

```
<pre><code>if (uttryck)
 sats;
else {
 en eller flera satser
}</code></pre>
<hr class="separator"/>
<pre><code>if (uttryck) {
 en eller flera satser
}
else
 sats;</code></pre>
```

The message from the validator is: "element 'pre' not allowed yet; missing required element 'p'". I think it is good not to overuse this separator, perhaps the separator in the example and should be solved with CSS to mark the boundaries of the pre/code sections? Perhaps my example is not the intended use of the hr sep. Let me know what you think.

josteinaj:
My first thought is that hr would not be appropriate in this case, but I don't really have a strong opinion. In any case, it should be technically valid, so this is indeed a bug.

martinpub:
And I would agree with you @josteinaj :-) I guess the interpretation depends on how much we want our productions to preserve separation marks from the print editions. I think the main case is in text to indicate a thematic break. In my example, we had several code examples and the "thematic break" semantics is much weaker there, since we have rather a list of examples. But I think I agree, it should be technically valid. It could be an accompanying task for guidelines development to clarify more about the use.

oscarlcarlsson:
I have seen this happen in poems as well.

josteinaj:
@martinpub @kalaspuffar solved?

kalaspuffar:
To my knowledge, yes.

PR #526, opened 2022-10-28
Allow separator in code

kalaspuffar:
Trying to solve: https://github.com/nlbdev/nordic-epub3-dtbook-migrator/issues/521

### DIFF
--- a/src/main/resources/xml/schema/2020-1/nordic-html5.rng
+++ b/src/main/resources/xml/schema/2020-1/nordic-html5.rng
@@ -2444,7 +2444,12 @@
         <element name="code">
             <ref name="attlist.code"/> <!-- @id, @title, @xml:space, @xml:lang, @lang, @dir, @epub:type, @class -->
             <zeroOrMore>
-                <ref name="inline"/> <!-- em, strong, dfn, code, samp, kbd, span, abbr, math, a, img, br, q, sub, sup, bdo -->
+                <zeroOrMore>
+                    <choice>
+                        <ref name="inline"/> <!-- em, strong, dfn, code, samp, kbd, span, abbr, math, a, img, br, q, sub, sup, bdo -->
+                        <ref name="hr"/>
+                    </choice>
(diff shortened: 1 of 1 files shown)


src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -2444,7 +2444,12 @@
         <element name="code">
             <ref name="attlist.code"/> <!-- @id, @title, @xml:space, @xml:lang, @lang, @dir, @epub:type, @class -->
             <zeroOrMore>
-                <ref name="inline"/> <!-- em, strong, dfn, code, samp, kbd, span, abbr, math, a, img, br, q, sub, sup, bdo -->
+                <zeroOrMore>
+                    <choice>
+                        <ref name="inline"/> <!-- em, strong, dfn, code, samp, kbd, span, abbr, math, a, img, br, q, sub, sup, bdo -->
+                        <ref name="hr"/>
```
josteinaj, line 2450: #521 is about hr elements between code blocks. This would allow hr elements in an inline context.

Today it seems like hr elements are only allowed adjacent to a p element. See `<define name="p">`. I think this is because of how hr was marked up in DTBook, where we did it using a class on the p element, since there was no dedicated element for it.

Here in EPUB we could simply remove it from `<define name"p">` and add it to the list in `<define name="block">`. Then it would be allowed between codeblocks and verses.

josteinaj (changes requested):

kalaspuffar:
Hi @josteinaj 

I think I misunderstood the task. I've changed the rules to follow what you've written above. Now this code example will pass validation.
```
            <pre><code>
        public static void main(String[] args) {
            ...
        }
        </code></pre>
            <hr class="separator"/>
            <pre><code>
        public static void main(String[] args) {
            ...
        }
        </code></pre>
```

Best regards
Daniel

josteinaj (approved):

Issue #519, opened 2022-07-01
No check for aria-describedby linking in text extractions/image descriptions


martinpub:
The 2020-1 guidelines [require text extractions/image descriptions to be put in a child aside element inside the figure, with proper linking between the img element and the aside using aria-describedby](https://format.mtm.se/nordic_epub/2020-1#text-extraction-from-images). However, there seems to be no check for aria-describedby linking in the validator.

The example from the guidelines would currently still be valid if aria-describedby is omitted, which would result in missing accessibility semantics and potentially rendering/skippability issues in reading systems.

A fix should introduce the following checks:

1. That aria-describedby is present on the img element of the figure
2. Link integrity from the aria-describedby reference to a matching fig-desc aside in the same figure container

josteinaj:
I'm starting my vacation this weekend. I'll be working a bit during week 29 (July 18th - July 22nd), so if there's a PR by then, I can have a look at it and hopefully make a release. Otherwise I'll be back from vacation August 9th, so I can look more at it during August.

martinpub:
Thanks @josteinaj! I think the time frame is OK, most people will be on vacation, just wanted to log this for now. Enjoy your vacation!

josteinaj:
@martinpub @kalaspuffar solved?

kalaspuffar:
To my knowledge, yes.

PR #527, opened 2022-11-10
Check that the extracted text has a described by link

kalaspuffar:
Trying to solve https://github.com/nlbdev/nordic-epub3-dtbook-migrator/issues/519


src/main/resources/xml/schema/2020-1/nordic2020-1.sch
```diff
@@ -630,6 +630,18 @@
         </rule>
     </pattern>
 
+    <pattern id="epub_nordic_253_d">
+        <title>Rule 253d</title>
+        <p>Extracted text needs to have a reference of the image the text have been extracted from.</p>
+        <rule context="html:figure[tokenize(@class, '\s+') = 'image']">
+            <let name="figdesc-id" value=".//html:aside[tokenize(@class, '\s+') = 'fig-desc']/@id" />
+            <let name="aria-id" value=".//html:img/@aria-describedby" />
+            <assert test="count(.//html:aside[tokenize(@class, '\s+') = 'fig-desc']) = 0 or $figdesc-id = $aria-id">
```
josteinaj, line 639: If both attributes are missing or empty, I think this assetion will succeed? This assertion covers the second check described by Martin in #519, but the first check also needs to be implemented. That is, img elements in this context needs to have a non-empty aria-describedby attribute.

josteinaj (changes requested):

src/main/resources/xml/schema/2020-1/nordic2020-1.sch
kalaspuffar: Hi @josteinaj 

I'm not sure if I understand the finer details here, but if I look at https://format.mtm.se/nordic_epub/2020-1#text-extraction-from-images it seems like the aria-describedby is only required in the context that we have, a fig-desc (extracted text) element. And this rule will verify both that it's present and that it's equal to the description.

Is there a nuance I've missed and could you give me an example?

Best regards
Daniel

kalaspuffar:
Hi @josteinaj 

I've added the test code I ran in the old repository. I now have this repository setup to run the java test directly in this repository. We might want to remove the generated XSLT files, though, to ease the review process. I'll look into that.

Best regards
Daniel

kalaspuffar:
I'm sorry that this PR is a bit messy but future ones will be more clean.

josteinaj:
After thinking about it, this looks good. Are we ready to merge?

josteinaj (approved):

kalaspuffar:
Hi @josteinaj 

Yes, this PR is ready to merge. I will create another convenience PR as well that you might want to merge later.

Best regards
Daniel

kalaspuffar:
Hi @erikmtm

Do you want to review this as well?

Best regards
Daniel

josteinaj:
I'm merging this.

@erikmtm: if you have comments, we can always change it later :+1:

PR #529, opened 2022-11-11
Add repository links for some repositories that used http instead of https

kalaspuffar:
Hi @josteinaj 

I have a newer version of maven that don't want to run the original test script because some of the repositories in Pipeline runs over HTTP. 

So this PR will add https versions of these repositories.

Best regards
Daniel


josteinaj:
Is this the root POM? I thought it wasn't used? I can build the Dockerfile without it.

kalaspuffar:
Hi @josteinaj 

Yes and yes.

This is for us while developing so we can run `mvn test` and run the old test cases to verify that we haven't introduced an error.

Best regards
Daniel

josteinaj:
👍

PR #534, opened 2022-12-06
Fixing issue with parts and chapters being in differnt files.

kalaspuffar:
Hi @josteinaj 

In this PR I've updated the `list-heading-and-pagebreak-references.xsl` files in the client so we have the right input data to the other schema files. Not sure why these were diffing.

I also added a test case and changed the code to handle the occurrence when you have a part and chapter in different files. The logic was complicated there, and I don't see a reason for it. I also updated one of the test books as it was incorrect.

Last but not least, the test book X50525A was incorrect regarding aria-described-by for images. The tests must have failed before, but I have not seen them. So had to fix this, so the tests passed.

Best regards
Daniel

### DIFF
--- a/client/src/test/resources/valid2020/EPUB/C00000-10-chapter.xhtml
+++ b/client/src/test/resources/valid2020/EPUB/C00000-10-chapter.xhtml
@@ -108,7 +108,7 @@
         <link rel="prev" href="C00000-11-conclusion.xhtml"/>
     </head>
     <body epub:type="bodymatter chapter" id="poem-part-chapter">
-        <h1>Chapter in poem part</h1>
+        <h2>Chapter in poem part</h2>
         <p>Should be a separate file in EPUB.</p>
     </body>
 </html>
--- a/client/src/test/resources/valid2020/EPUB/C00000-10-chapter.xhtml
+++ b/client/src/test/resources/valid2020/EPUB/C00000-10-chapter.xhtml
@@ -107,8 +107,10 @@
         <link rel="prev" href="C00000-09-part.xhtml"/>
         <link rel="prev" href="C00000-11-conclusion.xhtml"/>
     </head>
-    <body epub:type="bodymatter chapter" id="poem-part-chapter">
-        <h2>Chapter in poem part</h2>
-        <p>Should be a separate file in EPUB.</p>
+    <body>
+        <section epub:type="bodymatter chapter" id="poem-part-chapter">
+            <h2>Chapter in poem part</h2>
+            <p>Should be a separate file in EPUB.</p>
+        </section>
     </body>
 </html>
--- a/client/src/test/resources/valid2020/EPUB/C00000-13-part.xhtml
+++ b/client/src/test/resources/valid2020/EPUB/C00000-13-part.xhtml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-model href="http://www.daisy.org/pipeline/modules/nordic/nordic-html5.rng"?>
+<?xml-model href="http://www.daisy.org/pipeline/modules/nordic/nordic2015-1.sch"?>
+<!DOCTYPE html>
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns:epub="http://www.idpf.org/2007/ops" xml:lang="en" lang="en" epub:prefix="z3998: http://www.daisy.org/z3998/2012/vocab/structure/#"
+    xmlns:nordic="http://www.mtm.se/epub/">
+    <head>
+        <meta charset="UTF-8" />
+        <title>Valentin Haüy - the father of the education for the blind</title>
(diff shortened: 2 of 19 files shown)


client/src/test/resources/valid2020/EPUB/C00000-10-chapter.xhtml
```diff
@@ -108,7 +108,7 @@
         <link rel="prev" href="C00000-11-conclusion.xhtml"/>
     </head>
     <body epub:type="bodymatter chapter" id="poem-part-chapter">
-        <h1>Chapter in poem part</h1>
+        <h2>Chapter in poem part</h2>
         <p>Should be a separate file in EPUB.</p>
     </body>
```
josteinaj, line 113: Is this valid? I think all content documents are required to have a single `section` element as a child of the `body` element. So this should probably produce a validation error?

The 2020-1 guidelines says under "[3.1 Structural Divisons and Semantics](https://format.mtm.se/nordic_epub/2020-1/#structural-divisons-and-semantics)":

> The structural divisions of the publication are required to be semantically inflected by wrapping every structural part of the main text body in <section> elements, with proper nesting of subsections.


This is based on the Daisy KB. It mentions parts here (possibly in other articles as well): https://kb.daisy.org/publishing/docs/html/sections.html#faq

I think it should be:

```html
    <body>
        <section epub:type="bodymatter chapter" id="poem-part-chapter">
            <h2>Chapter in poem part</h2>
            <p>Should be a separate file in EPUB.</p>
        </section>
    </body>
```

client/src/test/resources/valid2020/EPUB/C00000-13-part.xhtml
```diff
@@ -0,0 +1,18 @@
...
+        <meta charset="UTF-8" />
+        <title>Valentin Haüy - the father of the education for the blind</title>
+        <meta content="C00000" name="dc:identifier" />
+        <meta name="viewport" content="width=device-width" />
+        <link rel="stylesheet" type="text/css" href="css/accessibility.css"/>
+    </head>
+    <body epub:type="bodymatter part" id="test-part">
+        <h1>Part title</h1>
+        <p>Part testing in different files</p>
+    </body>
```
josteinaj, line 17: Same here

client/src/test/resources/valid2020/EPUB/C00000-14-chapter.xhtml
```diff
@@ -0,0 +1,18 @@
...
+        <meta charset="UTF-8" />
+        <title>Valentin Haüy - the father of the education for the blind</title>
+        <meta content="C00000" name="dc:identifier" />
+        <meta name="viewport" content="width=device-width" />
+        <link rel="stylesheet" type="text/css" href="css/accessibility.css"/>
+    </head>
+    <body epub:type="bodymatter chapter" id="test-chapter">
+        <h2>Chapter title</h2>
+        <p>Chapter testing in different files</p>
+    </body>
```
josteinaj, line 17: Same here

client/src/test/resources/valid2020/EPUB/nav.ncx
```diff
@@ -175,19 +175,19 @@
...
+				<text>Rearmatter</text>
+			</navLabel>
+			<content src="C00000-07-rearnotes.xhtml#rearnotes_body"/>
+			<navPoint id="navPoint-27" playOrder="27">
+				<navLabel>
+					<text>End notes</text>
+				</navLabel>
+				<content src="C00000-07-rearnotes.xhtml#endnote_section"/>
+			</navPoint>
+		</navPoint>
```
josteinaj, line 190: Tab indented?

client/src/test/resources/valid2020/EPUB/nav.ncx
```diff
@@ -290,6 +290,18 @@
...
+				<text>Part title</text>
+			</navLabel>
+			<content src="C00000-13-part.xhtml#test-part"/>
+            <navPoint id="navPoint-16" playOrder="16">
+                <navLabel>
+                    <text>Chapter title</text>
+                </navLabel>
+                <content src="C00000-14-chapter.xhtml#test-chapter"/>
+            </navPoint>
+		</navPoint>
```
josteinaj, line 304: Tab indented?

kalaspuffar:
Hi @josteinaj 

I answer here as it's more of a general observation than connected to anything you have reported. These test files were copied from an old book and then adopted to test specific things. We are not running tests on all files, meaning the incorrect ones would not trigger the error.

I've added sections to those files and then updated the indentation. The tests still pass, so I think the code is still valid, and the test case should be correct. But I copied these files before we had a proper test book available from MTM.

Maybe I need to do a retake on the tests and update with the latest iteration of the test book in the validator, or perhaps even use the files in the repository instead, as we now have them in the same repository.

Best regards
Daniel

josteinaj (approved):

Issue #522, opened 2022-10-18
Further development, end of 2022 ("phase one")


josteinaj:
- [x] Finalize and release an official version (without Pipeline 2)
    - [x] support validation according to the 2020-1 EPUB guidelines (required)
    - [x] support validation according to the 2015-1 EPUB guidelines (optional, if not too much work)
    - [x] when MTM and NLB are happy with the version: create a tag and build that tag on docker hub
    - [x] add documentation about how to use the new version in the gh-pages branch (keep the documentation for how to run the old version, for those who need to convert between EPUB, HTML and DTBook)
- [x] Fix a few remaining issues with the RelaxNG/Schematron files
    - [x] #478 (see also: #509)
    - [x] #520
    - [x] #521
    - [x] #519
- [x] #395
    - [x] Verify that this fixes: #515
    - [x] Verify that this fixes: #512
- [x] Run in Docker with a REST API (not the Pipeline 2 Web API)
- [x] Run from command line (not the Pipeline 2 CLI)
- [x] Code should be in this official repo: [nlbdev/nordic-epub3-dtbook-migrator](https://github.com/nlbdev/nordic-epub3-dtbook-migrator)
- [x] Possible to run with or without Ace and Epubcheck
    - [x] #506
    - [x] Re-implement support for Epubcheck validation (currently requires Pipeline 2)

josteinaj:
@kalaspuffar I added two items under "Finalize and release an official version (without Pipeline 2)". We have touched on this on e-mail earlier, but it wasn't described here. We haven't finalized and released an official version until we have documentation and a tagged version on docker hub. I hope this is ok.

josteinaj:
The API is the same as for https://github.com/mtmse/talking-book-validator/, right? Is there any documentation for that one?

@martinpub @karladamt @oscarlcarlsson what do you think about documentation? Also, could you test the latest version, and see that it works for you as a replacement for your talking-book-validator? It's available as the docker image `nlbdev/nordic-epub3-dtbook-migrator:latest`.

kalaspuffar:
Hi @josteinaj 

The current documentation we use today is the
https://github.com/nlbdev/nordic-epub3-dtbook-migrator/blob/master/web/docs/api_v1.yaml

This file explains the API in an OpenAPI format that you could generate a live document/viewer and that could be used for testing and verification as well. But it should also be read as a simple file.

For more information about the format you could read up on https://www.openapis.org/

Best regards
[...]

PR #539, opened 2022-12-20
Handle the case when you have chapters that only have an aria label and no heading.

kalaspuffar:
Hi @josteinaj 

I think this is the issue vaguely referred to in https://github.com/nlbdev/nordic-epub3-dtbook-migrator/issues/522.

Currently, we will error and say that chapters are in the wrong level if we don't have a heading giving it a specific value for the level. When you have chapters that only have a section `aria-label`, we can't have any opinion on the specific placement in the hierarchy.

My fix might be a bit minimal, but if the XSLT doesn't create a `@data-heading-depth`, then we probably only have an `aria-label` section. Is that enough of a check, or do I need to see if we don't have an `aria-label-by`, then we need some work on the XSLT too?

Best regards
Daniel


src/main/resources/xml/schema/2020-1/nordic2020-1.nav-references.sch
```diff
@@ -125,9 +125,9 @@
             <let name="document-in-nav" value="((preceding::html:a | self::*) intersect ancestor::html:nav//html:a)[substring-before(@href,'#') = $href][1]"/>
             <let name="document-in-nav-depth" value="count($document-in-nav/ancestor::html:li)"/>
             <let name="depth-in-nav" value="count(ancestor::html:li)"/>
-            <let name="depth-in-content" value="$result-ref-first/xs:integer((@data-heading-depth, @data-sectioning-depth)[1])"/>
```
josteinaj, line 128: First a thing I notice now about `@data-sectioning-depth`, that might cause problems:

If the part is contained in a separate content file (which I think is the most common case), then we should add one to `data-sectioning-depth`. So: find the first first preceding `c:result` with `@data-heading-element='h1'`. If it is a part according to either `@data-document-epub-type` or `@data-document-role`, then add one to `data-sectioning-depth`. This assumes that the part has a `h1` element, which might not be a good assumption? But if we assume that it does have a `h1`, we could do something like:

```xml
<let name="data-sectioning-depth"
     value="$result-ref-first/xs:integer(@data-sectioning-depth)
            + if ((preceding::c:result[@data-heading-element='h1'])[1][
                  @data-document-epub-type = 'part'
                  or @data-document-role = 'doc-part'
            ]) then 1 else 0"/>
```

A better solution might be to do a `p:wrap-sequence` before `<p:for-each name="epub3-validate.step.group-nav-references.iterate-html">` in `…/xproc/step/2020-1/epub3-validate.step.xpl`. Then the whole book would be in the context before running `list-heading-and-pagebreak-references.xsl` so that the `@data-sectioning-depth` could be correctly calculated there. If performance becomes a problem when wrapping all the content, we could `p:delete` all irrelevant elements before wrapping and running the XSLT.

src/main/resources/xml/schema/2020-1/nordic2020-1.nav-references.sch
```diff
@@ -125,9 +125,9 @@
...
+            <let name="depth-in-content" value="$result-ref-first/xs:integer(@data-heading-depth)"/>
 
-            <assert test="not($result-ref-first) or $depth-in-nav = $depth-in-content">[nordic_nav_references_2b] The nesting of headlines in the content does not match the
+            <assert test="not(@data-heading-depth) or not($result-ref-first) or $depth-in-nav = $depth-in-content">[nordic_nav_references_2b] The nesting of headlines in the content does not match the
```
josteinaj, line 130: By removing `@data-sectioning-depth`, there won't be any errors for sectioning elements with `aria-label`. I think we should validate `aria-label` headings as well.

And I think it's best to split the assertion into two:

- the depth in the navigation document should be the same as the sectioning depth in the content documents
- if there is a heading element, then the heading element depth should be the same as the sectioning depth, for depths up to 6. For sectioning depths deeper than 6, the heading element depth must be 6.

src/main/resources/xml/schema/2020-1/nordic2020-1.nav-references.sch
kalaspuffar: Hi @josteinaj 

Thank you for the example. It needed some minor changes and a lot of testing but the general concept was helpful.

I don't know if the `data-document-role` will be set and if so could be multiple values. If not this should work just fine.

Best regards
Daniel

kalaspuffar:
Hi @josteinaj 

I have made the first part of the changes to this document. I wonder if we need to handle a depth larger than 6. I need to see an example to test this further. Maybe we should postpone that change until we have a document with significant depth.

Please review again.

Best regards
Daniel

src/main/resources/xml/schema/2020-1/nordic2020-1.nav-references.sch
```diff
@@ -125,9 +125,14 @@
             <let name="document-in-nav" value="((preceding::html:a | self::*) intersect ancestor::html:nav//html:a)[substring-before(@href,'#') = $href][1]"/>
             <let name="document-in-nav-depth" value="count($document-in-nav/ancestor::html:li)"/>
             <let name="depth-in-nav" value="count(ancestor::html:li)"/>
-            <let name="depth-in-content" value="$result-ref-first/xs:integer(@data-heading-depth)"/>
 
-            <assert test="not(@data-heading-depth) or not($result-ref-first) or $depth-in-nav = $depth-in-content">[nordic_nav_references_2b] The nesting of headlines in the content does not match the
+            <let name="data-part-depth"
+                 value="if ($result-ref-first/preceding-sibling::c:result[@data-heading-element='h1'][1]/
+                  tokenize(@data-document-epub-type, '\s+') = 'part' or @data-document-role = 'doc-part') then 1 else 0" />
```
josteinaj, line 131: This will work in most cases. But if a part has a `aria-label` heading, then it won't work. I think we need to handle those cases as well.

kalaspuffar:
Hi @josteinaj 

I slightly changed the xlst to get the section type of the document so we can check with the previous one. If the current document is a chapter with a previous part, we will add one to the section depth counter.

I hope this is a more robust solution. If I understand the specification, the `epub:type` attribute is mandatory, and the `role` attribute shall be used if a suitable role is available.

Please review it again.

Best regards
Daniel

josteinaj:
I'll have to look more at this in January, I only have a couple of hours left until my Christmas vacation.

I need to check if we handle cases where a part does not have a visible headline (i.e. it uses `aria-label`), and where there's a chapter with a `aria-label` headline that follows a part, but is not in a part. In that corner case we might actually not be able to determine the depth based only on the content documents, but have to rely on the navigation document to determine if a section is in a part or not. We might also have to check for the bodymatter/backmatter boundary, we'll see.

kalaspuffar:
Hi @josteinaj 

Do we have any update on this PR?

Best regards
Daniel

josteinaj:
Hi @kalaspuffar.

Sorry, I haven't had time to look more into this.

I suspect we need more tests for this?

kalaspuffar:
Hi @josteinaj 

Well, I'll list the ones we have, and you decide if I should create another one.

1. X50525A-12 is testing when we have parts and chapters in the same file with the correct heading levels.
2. C0000-09 and 10 are testing chapters and parts in different files with the correct heading levels.
3. C0000-13, 14, and 15 are testing when a chapter does not have a heading, only an aria-label
4. C0000-16 and 17 are testing when the part has no heading, just an aria-label, and the chapter has the correct heading level.

Best regards
Daniel

josteinaj:
Ok, thank you. I'll look more at the tests later this week. But if I don't get time this week (which is likely), let's just merge it early next week and I'll assume it's fine.

PR #546, opened 2023-03-01
Adding PEF validation.

kalaspuffar:
Hi  @josteinaj 

I've done some work to add PEF validation to the validator. I've not gotten access to the newer RNG file, so I ran this against the old one. But when I get the new one, I will make another PR to replace that one.

With this change, we slowly move against a 1.0.0 release of the client. Still missing DTBook validation, at least.

I also think I figured out a way to make releases so we can bump the version number for the client and get the output a little bit more readable when we get a report. Always nice to know which version made the validation.

Best regards
Daniel


josteinaj:
Hi @kalaspuffar.

Validation of Nordic DTBook according to the 2015 guidelines is not a requirement, but would of course be nice. So we shouldn't need to wait for DTBook validation before making a release.

I've not heard about PEF validation until now. Is it just a generic PEF validation, or are someone working on nordic PEF guidelines? PEF is not mentioned as part of phase one: https://github.com/nlbdev/nordic-epub3-dtbook-migrator/issues/522

kalaspuffar:
Hi @josteinaj 

No, this is part of Phase two initiated by MTM. These PEF validation rules are the generic first version. There is a newer one that I have not gotten access to yet.

I'm currently looking at validation for Daisy 2.02, Daisy 3, and DTbook (both zipped and standalone). 

We can release the client with version 0.6.0 before that. But I was just talking about what initially was thinking could be version 1.0 when writing the validation client.

Best regards
Daniel

josteinaj:
Ok. On the one hand it's nice and useful to have validation of these formats, but I'm worried about the added complexity. I will have a discussion with MTM before we merge this.

josteinaj:
CC @oscarlcarlsson @karladamt

Issue #547, opened 2023-03-08
Heading level check incorrect outside of bodymatter when there are parts


keyvantextalk:
There's an issue where EPUB nav.xhtml files which refer to pages outside of the bodymatter are marked as having an incorrect level. This only occurs in publications with parts and where the page has no explicit heading (ie. we rely on the aria-label)

> [nordic_nav_references_2b] The nesting of headlines in the content does not match the nesting of headlines in the navigation document. The toc item `(<a href="nordic_epub_2020_1_-053-backmatter.xhtml#d5e9879">)` in the navigation document is not nested at the correct level. The referenced document (nordic_epub_2020_1_-053-backmatter.xhtml) occurs in the navigation document at nesting depth 1 (it is not contained inside other sections such as a part or a chapter). The referenced headline (nordic_epub_2020_1_-053-backmatter.xhtml#d5e9879) occurs in the navigation document at nesting depth 1 (it is not contained inside other sections such as a part or a chapter). The referenced headline (`<>[aria label content]</>) occurs in the content document nordic_epub_2020_1_-053-backmatter.xhtml as a `` which implies that it should be referenced at nesting depth 2 in the navigation document.

PR #548, opened 2023-03-08
Backmatter hierarchy 

kalaspuffar:
Fixing issues when we look for chapters without heading that aren't chapters but backmatter, for instance.

This fixes https://github.com/nlbdev/nordic-epub3-dtbook-migrator/issues/547


kalaspuffar:
Hi @josteinaj 

As this is a small issue you might be able to look at this. Still waiting for a response.

Best regards
Daniel

PR #549, opened 2023-03-14
Updated to use Saxon 11 and update all schemas to XSLT 3.0

kalaspuffar:
Hi @josteinaj 

We found that some books that MTM ran crashed EPUBCheck. Some functions of Saxon were not available in version 9. New EPUBCheck 5.0.0 uses Saxon 11. And it required us to update all the schemas to version XSLT 3.0. 

XSLT 2.0 is no longer supported in Saxon 11.

Also needed to update Rule 1 of the `nordic2015-1.opf-and-html.sch` as it did not support the setting of global variables. I have tested and verified that the new solution finds ids that aren't unique.

Best regards
Daniel

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -10,23 +10,23 @@ Create a new version of the tooling with the current schema you
 need to one dependency to convert the schema files to xslt.
 More over you need to download and install saxon-xslt.
 
-#### Download schxslt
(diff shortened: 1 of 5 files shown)


josteinaj:
Sounds good 👍

Issue #536, opened 2022-12-13
Allow arbitrary order of header elements.


kalaspuffar:
Hinderburg creates epubs that have an order for the elements that is not allowed.

```
 <head>
    <meta charset="UTF-8" />
    <title>Bibliotekariestereotypen i populärkulturen</title>
    <meta name="dc:title" content="Bibliotekariestereotypen i populärkulturen" />
    <meta name="dc:creator" content="Richard Ohlsson" />
    <meta name="dc:identifier" content="V002631" />
    <meta name="dc:format" content="ePub 3.0" />
    <meta name="dc:publisher" content="MTM" />
    <meta name="ncc:generator" content="Narrator Studio 1.50.2450" />
    <meta name="viewport" content="width=device-width" />
    <link rel="stylesheet" type="text/css" href="style.css" />
  </head>
```

The order in the beginning of the file should be:

```
    <meta charset="UTF-8" />
    <title>Bibliotekariestereotypen i populärkulturen</title>
    <meta name="dc:identifier" content="V002631" />
    <meta name="viewport" content="width=device-width" />
```

After that, we allow any or none of the meta or style link items.

The problem is that we can't interleave elements using RNG if they are the same element. So we can't interleave 3 meta elements as they aren't unique.

kalaspuffar:
Hi @josteinaj 

How do we handle this? Remove the RNG validation and rewrite it in Schematron?

Or do you have a better solution here?

Best regards
Daniel

josteinaj:
Could it be done with a `<choice>` in RNG?

josteinaj:
Also, is this the 2015-1 guidelines?

kalaspuffar:
Hi @josteinaj 

I'm not sure if we can do it with choice. Then you need to say choose any of these and we need 4 of them. But we could have duplicates and then one could be missing.

Best regards
Daniel

kalaspuffar:
> Also, is this the 2015-1 guidelines?

No clue what Hindenburg creates, never worked with it.

josteinaj:
Ok. I haven't actually used Hindenburg myself, but as far as I know: Hindenburg does as little as possible to the EPUB. So the EPUB you put in, will be the EPUB you get out, except for the additon of media overlay and some other changes. I don't think it forces the output into the nordic markup guidelines.

[...]

PR #553, opened 2023-04-12
Allow unordered metadata

kalaspuffar:
Hi @josteinaj 

Trying to solve https://github.com/nlbdev/nordic-epub3-dtbook-migrator/issues/536

This is a small change moving the uniqueness check of the metadata fields to Schematron.

I created a new rule called nordic_1. Maybe this number is already used, then we will change it, but validating unique fields in any order is not allowed in RelaxNG.

I did a quick check if I could add the Schematron rules as an extension of the RelaxNG rules. Seemed not to work so to be careful I created a new rule.

### DIFF
--- a/src/main/resources/xml/schema/2020-1/nordic2020-1.sch
+++ b/src/main/resources/xml/schema/2020-1/nordic2020-1.sch
@@ -14,10 +14,10 @@
         <title>Rule 1</title>
         <p>Only allow one charset, identifier, title and viewport</p>
         <rule context="html:head">
-            <assert test="count(html:meta[@charset='UTF-8'])=1">[nordic01] charset seems not to be in head exactly once</assert>
-            <assert test="count(html:meta[@name='dc:identifier'])=1">[nordic01] dc:identifier seems not to be in head exactly once</assert>
-            <assert test="count(html:title)=1">[nordic01] title seems not to be in head exactly once</assert>
-            <assert test="count(html:meta[@name='viewport'])=1">[nordic01] viewport seems not to be in head exactly once</assert>
+            <assert test="count(html:meta[@charset='UTF-8'])=1">[nordic01] charset must be in head exactly once</assert>
+            <assert test="count(html:meta[@name='dc:identifier'])=1">[nordic01] dc:identifier must be in head exactly once</assert>
+            <assert test="count(html:title)=1">[nordic01] title must be in head exactly once</assert>
+            <assert test="count(html:meta[@name='viewport'])=1">[nordic01] viewport must be in head exactly once</assert>
         </rule>
     </pattern>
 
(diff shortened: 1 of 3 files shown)


src/main/resources/xml/schema/2020-1/nordic2020-1.sch
```diff
@@ -10,6 +10,17 @@
...
     <ns prefix="prism" uri="http://www.prismstandard.org/specifications/3.0/PRISM_CV_Spec_3.0.htm#"/>
 
+    <pattern id="epub_nordic_1">
+        <title>Rule 1</title>
+        <p>Only allow one charset, identifier, title and viewport</p>
+        <rule context="html:head">
+            <assert test="count(html:meta[@charset='UTF-8'])=1">[nordic01] charset seems not to be in head exactly once</assert>
+            <assert test="count(html:meta[@name='dc:identifier'])=1">[nordic01] dc:identifier seems not to be in head exactly once</assert>
+            <assert test="count(html:title)=1">[nordic01] title seems not to be in head exactly once</assert>
+            <assert test="count(html:meta[@name='viewport'])=1">[nordic01] viewport seems not to be in head exactly once</assert>
```
josteinaj, line 20: @kalaspuffar could you rephrase these to be "must be" instead of "seems not to be"? Just for consistency with other assertions.

Also, I don't think you need to require at least one title element and charset meta element, since I don't think they are required by the nordic standard.
- Missing title elements should ideally be picked up by epubcheck and/or Ace.
- And for missing charset; I guess UTF-8 would be implicit? I have not checked.

josteinaj (changes requested):

src/main/resources/xml/schema/2020-1/nordic2020-1.sch
kalaspuffar:
```
2.5.2 Title
The <title> element of every xhtml content document must match the dc:title metadata of the package file.

2.5.3 Metadata
The content documents are required to contain two <meta> elements.

<meta name="dc:identifier" content="_[production UID]_"/>
<meta name="viewport" content="width=device-width"/>
The production UID must match the dc:identifier in package.opf.
```

You are correct that there is no mention of either title or charset. But they were validated in the original rules so I didn't want to remove them when I changed to validate using Schematron. But if we feel confidable with removing them, then we could do it.

Not supplying a title or charset is something I would strongly not recommend, as they will ensure some structure. And I believe that EPUBCheck will even fail if you supply any other value for charset than UTF-8.

kalaspuffar:
```
[ace] Document does not have a non-empty &lt;title&gt; element<br/><br/>&lt;html xmlns=&quot;http://www.w3.org/1999/xhtml&quot; xmlns:epub=&quot;http://www.idpf.org/2007/ops&quot; xml:lang=&quot;en&quot; lang=&quot;en&quot;&gt; (media_overlay_test-001-foreword.xhtml)
[epubcheck] Line: -1 Col: -1 Error while parsing file: assertion failed: WARNING: The &quot;head&quot; element should have a &quot;title&quot; child element. (EPUB/media_overlay_test-001-foreword.xhtml)
[nordic01] charset seems not to be in head exactly once (EPUB/media_overlay_test-001-foreword.xhtml)
[nordic01] title seems not to be in head exactly once (EPUB/media_overlay_test-001-foreword.xhtml)
```

I've rerun it, and the title is validated with the EPUBCheck and the Ace rules, but they do not require the charset.

Issue #540, opened 2022-12-20
Titles are validating even without <!DOCTYPE html>


oscarlcarlsson:
I just notices that files turn out as valid even though they are missing <!DOCTYPE html> in the validator. I believe that this has been an issue for a long time, since several of our enriched files are missing it.
According to 2.5.1.2 Document Type Declaration in the 2020-1 guidelines, this is required.
Is this left out on purpose?

josteinaj:
Checking the doctype (or the XML declaration) is not possible using RelaxNG og Schematron. The validator that used XProc and Pipeline 2 used Java to extract the doctype and XML declaration, but also had a fallback to a pure XProc/XSLT implementation (although it was less efficient).

@kalaspuffar do you know if the doctype validation was copied over from the old code (or reimplemented)?

kalaspuffar:
Hi.

We have added a card for this in Trello and will address it in the Prio order.

Best regards
Daniel

oscarlcarlsson:
I just had a look at some files where <!DOCTYPE html> was not included. and they still pass through the validator. Any thoughts on if this should be allowed or not?

josteinaj:
The XML and DOCTYPE declarations are required in the guidelines, so they should also be required in the validator. Both for 2015-1 and 2020-1.

PR #559, opened 2023-04-24
Checking for the presence of doctype.

kalaspuffar:
Hi @josteinaj 

In this PR, I try to check the presence of doctype for all the content files of the EPUB.

This will hopefully solve https://github.com/nlbdev/nordic-epub3-dtbook-migrator/issues/540

Best regards
Daniel

### DIFF
--- a/client/src/main/java/org/daisy/validator/Util.java
+++ b/client/src/main/java/org/daisy/validator/Util.java
@@ -102,6 +102,9 @@ public class Util {
         String fileContent = "";
         while ((line = br.readLine()) != null) {
             fileContent += line;
+            if (fileContent.matches(".*<[a-zA-Z].*")) {
+                return false;
+            }
             if (fileContent.contains("<!DOCTYPE html>")) {
                 return true;
             }
--- a/client/src/main/java/org/daisy/validator/Util.java
+++ b/client/src/main/java/org/daisy/validator/Util.java
@@ -4,6 +4,7 @@ import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileOutputStream;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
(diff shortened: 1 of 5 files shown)


client/src/main/java/org/daisy/validator/Util.java
```diff
@@ -95,6 +96,19 @@ public static String getRelativeFilename(String currentfile, String link) throws
...
+    public static boolean checkIfContainsDoctype(InputStream inputStream) throws Exception {
+        BufferedReader br = new BufferedReader(new InputStreamReader(inputStream));
+        String line;
+        String fileContent = "";
+        while ((line = br.readLine()) != null) {
+            fileContent += line;
+            if (fileContent.contains("<!DOCTYPE html>")) {
+                return true;
+            }
+        }
```
josteinaj, line 108: I think the doctype should be before the root element. What do you think?

You could for instance split each line on `<` and:

1. if the first letter is `?`, then continue (probably the XML declaration)
2. if the string starts with `!DOCTYPE html>` => `return true;`
3. if the first letter is `!`, then continue (probably a comment)
4. if the split position is 1 or more => this is an element => `return false;`

I think that would work. Or maybe use a regex. Or something else :)

josteinaj (changes requested):
Looks good, but I wonder if we should make sure that the doctype is before the root element. See comment.

You could probably have a quite similar implementation for checking the XML declaration.

client/src/main/java/org/daisy/validator/Util.java
kalaspuffar: Hi @josteinaj 

Sounds reasonable. A simple regex just checking if we have a `<` followed by a letter should be enough to shortcircuit the execution and improve execution, ensuring that the doctype is before the actual document. 

I also updated the test case and moved the doctype into the document.

Best regards
Daniel

client/src/main/java/org/daisy/validator/Util.java
```diff
@@ -102,6 +102,9 @@ public static boolean checkIfContainsDoctype(InputStream inputStream) throws Exc
         String fileContent = "";
         while ((line = br.readLine()) != null) {
             fileContent += line;
+            if (fileContent.matches(".*<[a-zA-Z].*")) {
```
josteinaj, line 105: A corner case here would be if there's no newline after the doctype. So this would fail this check:

```html
<!DOCTYPE html><html>
<head></head><body></body></html>
```

However, I think it's good enough. It's nicer to have newlines after the doctype anyway I think.

josteinaj (approved):

josteinaj:
I see there's a conflict in TestValid.java that prevents merging. Feel free to merge when it's resolved.

PR #562, opened 2023-06-05
Allow hr in poems

kalaspuffar:
Hi @josteinaj 

We have seen the usage of Horizontal rules in books within poems to mark essential breaks in the flow. I've made a small change to the RNG to allow for this.

Best regards
Daniel

### DIFF
--- a/client/src/main/java/org/daisy/validator/EPUBFiles.java
+++ b/client/src/main/java/org/daisy/validator/EPUBFiles.java
@@ -616,6 +616,6 @@ public class EPUBFiles {
     }
 
     public String getPackageFile() {
-        return this.packageOBF;
+        return this.packageOPF;
     }
 }
--- a/client/src/main/java/org/daisy/validator/Util.java
+++ b/client/src/main/java/org/daisy/validator/Util.java
@@ -90,9 +90,12 @@ public class Util {
     );
 
     public static String getRelativeFilename(String currentfile, String link) throws IOException {
+        if (new File(currentfile).getParentFile() == null) {
+            return link;
+        }
         String parentPath = new File(currentfile).getParentFile().getAbsolutePath();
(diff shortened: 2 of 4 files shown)


kalaspuffar:
Hi @josteinaj 

I saw two more file changes, those a bugfixes for the flow. I thought we already had handled them.

Best regards
Daniel

kalaspuffar:
Hi.

Added a test case as well to ensure functionality.

Best regards
Daniel

josteinaj:
Hi.

While the code looks good, this is not mentioned as an option in the guidelines:
https://format.mtm.se/nordic_epub/2020-1/#poetry-and-verse

I think we need to release a version of the validator for 2020-1 before we start revising the guidelines.

martinpub:
Hi everyone. I agree with @josteinaj, even though there is a strong case for using separators in verse markup, as it is common in poems.

However, an observation is that the "Poetry and verse" section in 2020-1 does not forbid hr thematic break, and that "provid[ing] distinguishable paragraph-level thematic breaks" ("3.5 Thematic Breaks in the Text Flow") through hr markup could be interpreted as applicable also in verse settings.

I don't see any other good ways of marking up poems with stanzas separated by e.g. asterisks to indicate thematic breaks in 2020-1 if this is disallowed.

FIY: me, @oscarlcarlsson and @AndersEkl have started planning for a guidelines revision, more info with a draft plan will be available soon.

AndersEkl:
Like @martinpub said, there is nothing in the 2020-1 guidelines that forbids the use of thematic breaks in a poetry and verse context. I would not object to the validator allowing it. And we can make it clearer in the upcoming revision that it is allowed.

karladamt:
@josteinaj it would help us a lot if we could merge this now, it is blocking some titles for us as we are transending to epub.

josteinaj (approved):

PR #563, opened 2023-06-30
Catch up

kalaspuffar:
Hi @josteinaj 

As we talked about in our meeting, we wanted to prepare for release. This PR is a bit of a catch-up for minor fixes that need to be in the release.

* Allow HR in Poems
* Visual fixes to the report
* Updates to the Docker script to enable deployment
* More information when an audio clip is outside of the audio file.
* Relative file bugfix
* Typo fix
* The ACE fixes we talked about, adding the list of rules and only use WCAG A -> AAA.
* I also added a location for the ACE validation issues so we could more easily locate the problems.

Best regards
Daniel


client/src/main/java/org/daisy/validator/ace/ACEValidator.java
```diff
@@ -151,4 +192,12 @@ public static String getVersion() {
     public static boolean isAvailable() {
         return version != null;
     }
+
+    public static void main(String[] args) {
+        List<Issue> list = validate(new File("/home/danielp/daisywork/epub/color_contrast.epub"));
```
josteinaj, line 197: @kalaspuffar could the file path be an argument instead of hard coded?

PR #565, opened 2023-08-14
Allow sentences in poems, verses and line-by-line constructions.

kalaspuffar:
Hi @josteinaj 

After producing books with sentence markup for MTM with poems, verse, and line-by-line constructions, we saw that adding classification for sentences was not allowed and would add extra unnessarary complexity to the format if we didn't do a minor change to the validator.

This will allow the lines of these constructions to also carry the sentence class for use in narrating systems.

Best regards
Daniel

### DIFF
--- a/src/main/resources/xml/schema/2020-1/nordic-html5.rng
+++ b/src/main/resources/xml/schema/2020-1/nordic-html5.rng
@@ -2289,9 +2289,11 @@
                             <value>line</value>
                             <value>line_indent</value>
                             <value>line_longindent</value>
-                            <value>sentence</value>
                         </choice>
                     </oneOrMore>
+                    <optional>
+                        <value>sentence</value>
+                    </optional>
                 </list>
             </attribute>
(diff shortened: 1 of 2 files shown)


src/main/resources/xml/schema/2020-1/nordic-html5.rng
```diff
@@ -2289,6 +2289,7 @@
                             <value>line</value>
                             <value>line_indent</value>
                             <value>line_longindent</value>
+                            <value>sentence</value>
```
josteinaj, line 2292: This would allow sentence without line. Could you move it to a separate `<optional>`?

josteinaj (changes requested):

src/main/resources/xml/schema/2020-1/nordic-html5.rng
kalaspuffar: Correct, that would be optional.

PR #568, opened 2023-11-16
Release 2.0.1

josteinaj:


josteinaj:
@kalaspuffar I've tagged and released v2.0.1 here. However, would it be possible for you to make it so that the pom file reads the version from the current git tag somehow? So we don't have to update `client/pom.xml` for every minor change? Alternatively make it so that the version in the pom is not logged or reported anywhere?

PR #570, opened 2024-03-20
Allow id, lang, title and direction in poems

kalaspuffar:
Hi @josteinaj 

When we created the new rule for poems in 2020, I think we were a bit too restrictive. Now, when we produce TTS material, we can't link the SMIL files to parts of a poem or line-by-line. I guess that this is an oversight.

Best regards
Daniel


josteinaj (approved):
Looks good 👍

PR #573, opened 2024-05-20
Fixing issue where books are longer than 99 hours.

kalaspuffar:
Hi @josteinaj 

We just received an error in a book with 120 hours of Economic forecasts. The regular expression only supported two numbers in the hour's value. I've raised this to three numbers, so hopefully, we never see a book with 1k hours.

Best regards
Daniel

### DIFF
--- a/client/src/main/java/org/daisy/validator/Util.java
+++ b/client/src/main/java/org/daisy/validator/Util.java
@@ -86,7 +86,7 @@ public class Util {
     }
 
     private static final Pattern milliPattern = Pattern.compile(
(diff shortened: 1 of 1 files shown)


kalaspuffar:
> Looks good. Do you think we could do a \d+ to avoid having an upper limit? Or could there be issues with that?

We could do that, but I remember that the search could be really slow when you do backward lookups, so I think this solution is safer and probably more performant. It was a while ago that I read up on this, so there might have been changes since then.

Best regards
Daniel

